PR #217, opened 2014-10-18
Removing ad support reference

jhuggart:
This reference caused my app to be flagged as using the IDFA. This forces you to choose why you are using the IDFA (I wasn't) when submitting to the app store.

### DIFF
--- a/Analytics/SEGAnalyticsUtils.m
+++ b/Analytics/SEGAnalyticsUtils.m
@@ -2,7 +2,6 @@
 // Copyright (c) 2014 Segment.io. All rights reserved.
 
 #import "SEGAnalyticsUtils.h"
-#import <AdSupport/ASIdentifierManager.h>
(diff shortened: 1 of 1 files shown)


naqi:
Can you tell us why this was added? I was blind sided by this by Apple when submitting an update.

reinpk:
hey @naqi so sorry about that! it's being recorded to be used for advertising attribution:
https://segment.com/docs/libraries/ios/quickstart/#step-4-submitting-to-the-app-store

naqi:
@reinpk Thanks for your reply. I kept debugging it and found out that it was actually GoogleAnalytics library which was the main problem. I ended up accepting the IDFA questions just because we want to use GoogleAnalytics and submitted the update.

reinpk:
got it, thanks for following up! hope all is well going forward.

PR #233, opened 2014-12-19
Bump version of Optimizely SDK

kmet31:
from 0.7.55 to 1.0.60


travisjeffery:
Optimizely just released a version that supports iOS 6 too and we just updated to that! Thanks though!

PR #238, opened 2015-01-06
Fix mapping of createdAt -> $created

fatuhoku:
See expected behaviour from documentation:

https://segment.com/docs/api/tracking/identify/#special-traits

Untested.


lambtron:
More information:

Segment’s `createdAt` made a `createdAt` field in Mixpanel, with the relevant date string that I set through Segment, even though it’s supposed to just update `$createdAt`.

http://imgur.com/ZWbak2I

In the image above, `createdAt` is created when call is made to Segment (yesterday), but it should have just updated the other field.

travisjeffery:
Thanks!

PR #240, opened 2015-01-13
Properly cast objc_msgSend calls

ayn:
See 2014 WWDC Session 417. If not casted, the compiler will complains with `Too many arguments to function call, expected 0, have 3`.


travisjeffery:
Thanks!

PR #243, opened 2015-01-14
updating Taplytics integration to support user attributes and analytics

jonathannorris:
I attempted to follow the same format from your other integrations. Let me know if the mapping makes sense.


travisjeffery:
yep lgtm!

PR #250, opened 2015-02-05
Forward location to Flurry. Closes #112

f2prateek:

PR #264, opened 2015-02-19
Add location to flurry

travisjeffery:
@f2prateek made the same fix to flurry to highlight some ways that https://github.com/segmentio/analytics-ios/pull/250 could have implemented a bit better.

for one we didn't need a separate method to set the default value, since if there is no object for say latitude then that value will be nil, and when any message is sent to nil it's also nil aka 0. kind of an objc behavior thing that can useful to know. 

but let's say the default was something other than 0, another way to do a default like that is `[location objectForKey:@"longitude"] ?: 4;`. where 4 would be the default.

we didn't need to add to this method: 
https://github.com/segmentio/analytics-ios/pull/250#discussion-diff-24203107 but if we did you would have wanted to write it as:

``` objective-c
- (float)floatFromDictionary:(NSDictionary *)dictionary key:(NSString *)key default:(NSNumber *)default {  
// ...
}
```
that way when you called it would have looked like:

``` objective-c 
[self floatFromDictionary:location key:@"latitude" default:0];
```

rather than 

``` objective-c
[self floatForKey:location :@"horizontalAccuracy" :0]
```

notice how in my example, the params are named.

### DIFF
--- a/Analytics/Integrations/Flurry/SEGFlurryIntegration.m
+++ b/Analytics/Integrations/Flurry/SEGFlurryIntegration.m
@@ -62,6 +62,16 @@
   if (age) {
     [Flurry setAge:[age intValue]];
   }
+  
+  NSDictionary *location = [traits objectForKey:@"location"];
+  
+  if (location) {
+    float latitude = [[location objectForKey:@"latitude"] floatValue];
+    float longitude = [[location objectForKey:@"longitude"] floatValue];
+    float horizontalAccuracy = [[location objectForKey:@"horizontalAccuracy"] floatValue];
+    float verticalAccuracy = [[location objectForKey:@"verticalAccuracy"] floatValue];
+    [Flurry setLatitude:latitude longitude:longitude horizontalAccuracy:horizontalAccuracy verticalAccuracy:verticalAccuracy];
+  }
 }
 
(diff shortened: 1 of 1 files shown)


Analytics/Integrations/Flurry/SEGFlurryIntegration.m
```diff
@@ -62,6 +62,16 @@ - (void)identify:(NSString *)userId traits:(NSDictionary *)traits options:(NSDic
   if (age) {
     [Flurry setAge:[age intValue]];
   }
+  
+  NSDictionary *location = [traits objectForKey:@"location"];
+  
+  if (location) {
+    float latitude = [[location objectForKey:@"latitude"] floatValue];
```
f2prateek: So what happens if `[location objectForKey:@"latitude"] ` is `nil`? `latitude` simply gets set to 0?

The way I was reading this (as in Java) was `[nil floatValue]` would throw a NullPointerException.

travisjeffery: yep it would be 0. `nil` is magical like that in objc. if it where NSNull * on the other hand then it'd be a problem. this post here covers it in a bit more detail: http://nshipster.com/nil/

f2prateek:
@travisjeffery do you know where the CI error is coming from? Running into it locally as well.

f2prateek:
Closes #264

Issue #271, opened 2015-03-06
Google Analytics integration uses `defaultTracker`


johnhaitas:
`SEGGoogleAnalyticsIntegration` uses `[[GAI sharedInstance] defaultTracker]`. This is a problem if a tracker has been initialized before `SEGGoogleAnalyticsIntegration`'s `-start` method is called. This can be resolved by `SEGGoogleAnalyticsIntegration` maintaining its own tracker instance property and using that instead of `[[GAI sharedInstance] defaultTracker]`.

johnhaitas:
pull request here https://github.com/segmentio/analytics-ios/pull/272/files

PR #272, opened 2015-03-06
SEGGoogleAnalyticsIntegration: Replace usage of Google Analytics defaultTracker with instance property.

johnhaitas:
SEGGoogleAnalyticsIntegration uses [[GAI sharedInstance] defaultTracker]. This is a problem if a tracker has been initialized before SEGGoogleAnalyticsIntegration's -start method is called. This can be resolved by SEGGoogleAnalyticsIntegration maintaining its own tracker instance property and using that instead of [[GAI sharedInstance] defaultTracker].


johnhaitas:
this is to address this issue https://github.com/segmentio/analytics-ios/issues/271

f2prateek:
What happens if our users do want to call some method not supported by us on the tracker?

johnhaitas:
That `tracker` property could be made public if necessary or accessed in a different way.

What if your users want to use multiple trackers? (we are)

f2prateek:
I think using multiple trackers is likely a smaller use case than using a single one tracker

I'm not saying we should't support this feature, we just have to think about it a bit.

sperand-io:
closing in favor of #276. thanks!

PR #294, opened 2015-04-15
Swift cocoapods fix

tonyxiao:
Needed to make these changes to make Analytics work in a swift project that uses cocoapods.

### DIFF
--- a/Analytics.podspec
+++ b/Analytics.podspec
@@ -17,7 +17,7 @@ Pod::Spec.new do |s|
   s.requires_arc    = true
 
   s.subspec 'Core-iOS' do |ss|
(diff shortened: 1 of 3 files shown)


tonyxiao:
Ooops, just noticed a lot of whitespace changes. Sorry annoying editor.

f2prateek:
@tonyxiao Thanks for this, wouldn't we need to update the cocoapods version being used since swift support was only added 0.36.0 - or is that only for those who're directly using swift code in their project?

tonyxiao:
user who want to use this with swift will need to use cocoapods 0.36.0 or higher yes.

f2prateek:
Thanks!

PR #301, opened 2015-04-28
Segment-Kahuna Wrapper changes.

sharekris:


sperand-io:
Hey @sharekris thanks so much for the PR! :) This looks awesome.

@f2prateek can we find time to review before thursday?

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
```diff
@@ -38,42 +103,400 @@ - (void)validate {
...
+      [self addCompletedOrderElements:&attributes fromProperties:lowerCaseKeyProperties];
+    }
+    
+    // If we have collected any attributes, then we will call the setUserAttributes API
+    if  (attributes.count > 0) {
+      [KahunaAnalytics setUserAttributes:attributes];
+    }
+  }
+  @catch (NSException *exception) {
+    NSLog (@"Kahuna-Segment Exception : %@", exception.description);
```
f2prateek: What exception is this trying to catch?

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
```diff
@@ -38,42 +103,400 @@ - (void)validate {
...
+    [(*attributes) setValue:discount forKey:KAHUNA_LAST_PURCHASE_DISCOUNT];
+  } else {
+    [(*attributes) setValue:@0 forKey:KAHUNA_LAST_PURCHASE_DISCOUNT];
   }
 }
 
 - (void)screen:(NSString *)screenTitle properties:(NSDictionary *)properties options:(NSDictionary *)options {
-  // Track the screen view as an event.
-  [self track:SEGEventNameForScreenTitle(screenTitle) properties:properties options:options];
+  id trackAllPages = [self.settings objectForKey:@"trackAllPages"];
```
f2prateek: Can be simplified to:
`BOOL trackAllPages = [[self.settings objectForKey:@"trackAllPages"] booleanValue];`

Analytics/Integrations/Kahuna/SEGKahunaIntegration.h
```diff
@@ -13,3 +15,13 @@
 @property(nonatomic, copy) NSDictionary *settings;
 
 @end
+
+@interface KahunaPushMonitor : NSObject
```
f2prateek: Should this be prefixed with `SEG`?

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
```diff
@@ -19,16 +60,40 @@ - (id)init {
...
+        [KahunaAnalytics handleNotification:[KahunaPushMonitor sharedInstance].pushInfo withApplicationState:[KahunaPushMonitor sharedInstance].applicationState];
+        [KahunaPushMonitor sharedInstance].pushInfo = nil;
+      }
+      
+      [KahunaPushMonitor sharedInstance].kahunaInitialized = TRUE;
+    }
+    
+    [super start];
+  }
+  @catch (NSException *exception) {
```
f2prateek: What is the purpose of this catch block?

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
```diff
@@ -38,42 +103,400 @@ - (void)validate {
...
+        }
+      }
+    }
+    
+    // Track the attributes if we have any items in it.
+    if (attributes.count > 0) {
+      [KahunaAnalytics setUserAttributes:attributes];
+    }
+  }
+  @catch (NSException *exception) {
```
f2prateek: What is the purpose of this catch block?

f2prateek:
@sharekris Is there anything we should document on the Kahuna integration docs that came up during testing, particularly for the push notification features?

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
sharekris: All the Try-Catch are put in place as a safety mechanism for cases where the input dictionary keys or values change to un-anticipated types in the future or get modified to other object types that do not accept the called methods being made on them. Since this code once deployed cannot be pulled back with a new SDK release, this is purely added for defensive purposes.

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
sharekris: The extra checks have been added to ensure "trackAllPages" is indeed a NSNumber class and not something else that can possibly thrown an exception when booleanValue is called on the object.

Analytics/Integrations/Kahuna/SEGKahunaIntegration.h
sharekris: Yes. will make the change.

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
sharekris: Added for defensive purposes. Possibly an over-kill in this situation since we do check for the key to be a non-null string in the SDK anyway. Will remove the try-catch.

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
sharekris: Defensive in order to ensure all the key:values of the traits dictionary do not change to un-anticipated types.

sharekris:
@f2prateek There is nothing specific about push notifications to be added in the Getting Started docs link we shared with you. I am referring to http://mathb.in/34168?key=a52dc9efacd7066b1d5873f77e2d1a1581011141

The swizzled methods we put in place ensures we are able to get device tokens, device registration failures, push clicked etc. seamlessly behind the scenes without any further steps from the customer during an integration.

sharekris:
@f2prateek We have a new SDK 570 that we released a few days ago. We already updated our CocoaPods with this version. So when you get a chance can you update your CocoaPods to get the version 570 ? It is currently pointing to  SDK 556.  (https://github.com/sharekris/Specs/commit/0431a8c70d0a9249976e9fe2e38de52528dda6ac)

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
f2prateek: It will only ever be a boolean so we don't need this check.

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
f2prateek: I'm not sure I understand, could you explain with a concrete example of what might change in the future that this block will handle?

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
sharekris: OK. will remove the check.

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
sharekris: Since the traits dictionary is created by the Customer and passed to our wrapper layer, we have added a Try-Catch to ensure no unexpected crashes result. The dictionary's key:value while passing our type safety checks, can still result with an exception either due to sub-classing or class forwarding that customers or other SDKs can implement. Our main concern is having the App crash for bad data. 

Do let me know if you are not comfortable with the Try-Catch blocks and feel it is un-necessary with all the type safety checks we are doing and I can remove it.

f2prateek: If you just want to check the types, then I think you should just wrap the accessor method with the try catch, the current scope is too large and will hide any potential bugs in the implementation. In general, we prefer they be small in scope.

sharekris:
@f2prateek  I have removed TryCatch blocks around track: & identify: methods. In both cases the incoming dictionary is being asserted on and coerced into the right type, so no point adding a new layer of checks in the Kahuna wrapper.

f2prateek:
Thanks, rebased and merged into master.

PR #304, opened 2015-04-29
Add Apptimize Integration.

DHowett:
- SEGApptimizeIntegration.
- Copy the Apptimize headers, statically link the framework binary.

This required some Cocoapods calisthenics on our end (we used to ship a versioned directory, but that was a bit incongruous with the other provided SDKs.) and a couple Segment-specific APIs.


f2prateek:
@DHowett thanks, have you been able to do an end to end test with the Apptimize integration for iOS?

DHowett:
I managed to run tests and track events through Segment and receive them on the Apptimize dashboard.
User traits appeared to work as well.

f2prateek:
@DHowett Thanks, I should have clarified that I was interested in the visual optimizer particularly, and if there's any notes we should add in the public docs related to it.

DHowett:
Sorry about the misunderstanding.

The visual editor worked with some minor caveats: only views in controllers
that come onscreen after -[SEGApptimizeIntegration init...:] is called can
be styled consistently. It might be reasonable to move the Apptimize
initialization into a class initializer or load method, but I don't want to
do too much work if we aren't enabled.

PR #306, opened 2015-05-05
Fix for iOS 6 compatibility

philippec:
Reading the address of CBCentralManagerOptionShowPowerAlertKey inline fails
on iOS 6, use a temporary variable to compare.

### DIFF
--- a/Analytics/Helpers/SEGBluetooth.m
+++ b/Analytics/Helpers/SEGBluetooth.m
@@ -28,7 +28,10 @@ const NSString *SEGCentralManagerClass = @"CBCentralManager";
 
   _queue = dispatch_queue_create("io.segment.bluetooth.queue", NULL);
 
-  if (&CBCentralManagerOptionShowPowerAlertKey != NO) {
+  // Reading the address inline fails on iOS 6, must use a temporary
+  // variable to compare.
+  NSInteger addr = &CBCentralManagerOptionShowPowerAlertKey;
+  if (addr != 0) {
     _manager = [[centralManager alloc] initWithDelegate:self queue:_queue options:@{ CBCentralManagerOptionShowPowerAlertKey: @NO }];
   } else {
     _manager = [[centralManager alloc] initWithDelegate:self queue:_queue];


philippec:
With this change I am able to launch my application on an iOS 6 device.

Analytics/Helpers/SEGBluetooth.m
```diff
@@ -28,7 +28,10 @@ - (id)init {
 
   _queue = dispatch_queue_create("io.segment.bluetooth.queue", NULL);
 
-  if (&CBCentralManagerOptionShowPowerAlertKey != NO) {
+  // Reading the address inline fails on iOS 6, must use a temporary
+  // variable to compare.
+  NSInteger addr = &CBCentralManagerOptionShowPowerAlertKey;
+  if (addr != 0) {
```
f2prateek: Can we leave this as `addr != NO)`?

philippec: Given that addr is a NSInteger, I would recommend not.

philippec: But if this is what you prefer, I will amend my PR.

f2prateek: Sorry I meant why not change the temporary variable as a Boolean instead of messing with a pointer, since that's what CBCentralManagerOptionShowPowerAlertKey is: 

CBCentralManagerOptionShowPowerAlertKey

```
  BOOL show = CBCentralManagerOptionShowPowerAlertKey;
  if (show != NO) {
```

philippec: Ah! I see what you mean. No, we should not do what you have specified above. The check is to see if the symbol `CBCentralManagerOptionShowPowerAlertKey` exists, so checking for its address is the correct thing to do (the symbol does not exist on iOS6). I could do 

    BOOL show = &CBCentralManagerOptionShowPowerAlertKey;
    if (show != NO) {
    ...

but IMHO that obfuscates the meaning of getting the address. Does that make sense?

f2prateek: Indeed it does, thanks!

PR #307, opened 2015-05-06
Explicitly check if selector exists

philippec:
Simply checking the symbol address was not sufficient on iOS6.


philippec:
Following up to #306, it appears that in our Release build we were still trying to call `initWithDelegate:queue:options:` on iOS6. This explicit check makes sure that doesn't happen. We now have an App Store build that runs on iOS6.

PR #325, opened 2015-06-05
Taplytics Update: Added delayLoad and shakeMenu options and updated to latest version.

nemo:
Hey folks,

We've upgraded our SDK and added some options for starting up the SDK that a few clients were asking for.

The new options are:

- `delayLoad` (Number) – seconds to keep showing the splash screen until the experiments are loaded. Used for when running an experiment on the first page of the app.
- `shakeMenu` (Boolean) – If it should permanently disable the Shake Menu or not.

Thanks in advance! Let me know if this all looks good.

cc: @BayPhillips + @jonathannorris.


platedbay:
+1 for this!

Analytics/Integrations/Taplytics/SEGTaplyticsIntegration.m
```diff
@@ -94,6 +102,23 @@ - (NSString *)apiKey {
...
+    NSString *value = self.settings[@"delayLoad"];
+    
+    if (value != nil)
+        return [NSNumber numberWithInt:[value intValue]];
+    
+    return nil;
+}
+
+- (NSNumber *)shakeMenu {
+    NSString *value = self.settings[@"shakeMenu"];
```
f2prateek: I think the default serialization parses booleans as numbers so you'll have to do:

```obj-c
if ([(NSNumber *)[self.settings objectForKey:@"shakeMenu"] boolValue]) {
    return YES;
}
return NO;
```

Analytics/Integrations/Taplytics/SEGTaplyticsIntegration.m
```diff
@@ -28,9 +28,17 @@ - (id)init {
 }
 
 - (void)start {
-  [Taplytics startTaplyticsAPIKey:[self apiKey]];
+  NSDictionary *options = [[NSMutableDictionary alloc] init];
   
-  SEGLog(@"TaplyticsIntegration initialized with api key %@", [self apiKey]);
+  if ([self delayLoad])
+      [options setValue:[self delayLoad] forKey:@"delayLoad"];
```
f2prateek: Given that the Taplytics SDK is initialized asynchronously by the Segment SDK - does this setting actually work as mentioned when being used via Segment?

nemo: We believe so. Some initial tests showed that it worked internally. As long as the users end up initialize Segment in the `didFinishLaunching:` method.

We'll be testing this with a few clients over time and will modify as we go if there's any issues.

Analytics/Integrations/Taplytics/SEGTaplyticsIntegration.m
nemo: Oh cool, I'll edit that up right away.

Analytics/Integrations/Taplytics/SEGTaplyticsIntegration.m
f2prateek: Just to confirm it ok to to say `[options setValue:@0 forKey:@"delayLoad"];`?

nemo: Yup!

f2prateek:
Closed by #325

PR #326, opened 2015-06-08
Added delayLoad, and shakeMenu options for Taplytics. Removed random,…

f2prateek:
… old files in Tapstream's directory.

Upgraded Taplytics to latest version.

Made sure to grab shakeMenu as a number from Segment and turn it into a bool NSNumber for Taplytics.


Analytics/Integrations/Taplytics/SEGTaplyticsIntegration.m
```diff
@@ -94,6 +101,16 @@ - (NSString *)apiKey {
   return self.settings[@"apiKey"];
 }
 
+- (NSNumber *)delayLoad {
+    return (NSNumber *)[self.settings objectForKey:@"delayLoad"];
+}
+
+- (BOOL)shakeMenu {
```
nemo: Hey @f2prateek – we actually need shakeMenu to be an NSNumber with a value of @0 or @1. This will not work with our SDK. It looks like I may have sent the commit to my branch too late :(

f2prateek: Oops, thanks for catching that, I'll fix it.

nemo: No problem. Thanks for fixing!

Issue #278, opened 2015-03-23
CFBundleDisplayName


brianvoong:
SEGSegmentioIntegration.m

Line 77
NSDictionary *infoDictionary = [[NSBundle mainBundle] infoDictionary];

should be 
NSDictionary *infoDictionary = [[NSBundle mainBundle] localizedInfoDictionary];

otherwise it'll crash at 
Line 80
@"name": infoDictionary[@"CFBundleDisplayName"],

f2prateek:
I don't think the line numbers match up - https://github.com/segmentio/analytics-ios/blob/master/Analytics/Integrations/Segmentio/SEGSegmentioIntegration.m#L77.

Could you clarify what crash you are seeing as well?

brianvoong:
Oh, looks like I was on an old branch and the nil dictionary exception has been fixed to insert an empty string instead of nil.

In any case, I believe the current line on master:
https://github.com/segmentio/analytics-ios/blob/master/Analytics/Integrations/Segmentio/SEGSegmentioIntegration.m#L110

will get a dictionary that has an empty string for the constant CFBundleDisplayName for line:
https://github.com/segmentio/analytics-ios/blob/master/Analytics/Integrations/Segmentio/SEGSegmentioIntegration.m#L113

Using localizedInfoDictionary seems to be "better" for getting the app name.

f2prateek:
Cool, do you have any reference/docs why using localizedInfoDictionary might be better? Typically we try to keep analytics data indifferent to locale (so you don't end up with multiple app names for different regions).

besbes:
I'd suggest to revert this change and go back to using ``infoDictionary``. We're currently experiencing that on iOS 8.3, ``localizedInfoDictionary`` contains only the attributes that are actually localized in InfoPlist.strings, but does not include important attributes like ``CFBundleVersion`` and ``CFBundleShortVersionString``— in turn, this information is missing in our tracked events.

If you insist on using localized information, I'd suggest doing something like this:

    NSMutableDictionary *infoDictionary = [[[NSBundle mainBundle] infoDictionary] mutableCopy];
    [infoDictionary addEntriesFromDictionary:[[NSBundle mainBundle] localizedInfoDictionary]];

jklimaat:
I agree with @besbes' comment about the current behavior of `localizedInfoDictionary`.

[...]

PR #327, opened 2015-06-09
Use infoDictionary and override with values from localizedInfoDictionary

besbes:
This is related to issue #278


f2prateek:
Thanks!

PR #349, opened 2015-06-29
Localytics tests

f2prateek:


f2prateek:
failing because countly deleted their 3.0.0 tag

PR #355, opened 2015-07-06
UXCam SDK integration

richardgroves:
NB: Will not build as is due to a version problem with Countly - change their SDK version to 2.0.0 in integrations.json rather than 3.0.0 until they fix their CocoaPod.

Adding UXCam SDK to integrations.json
Adding SEGUXCamIntegration
Adding SEGUXCamIntegrationTests - very simple work in progress
Editing AnalyticsTest and settings.json files to add UXCam integration


f2prateek:
Committed the Podfile.lock in https://github.com/segmentio/analytics-ios/commit/189e0855920af6bc235c8dfa386694c1bf3dac40

richardgroves:
Thanks. Sorry about the PodFile.lock - am never sure if they should be committed or not.

PR #360, opened 2015-07-10
Add Apptimize root integration

f2prateek:


f2prateek:
@amillet89 This differs from the Android Apptimize root in 2 ways:

1. Events may be fired multiple times for a single experiment viewed 
2. Events will have the experimentId and variantId (their Android SDK doesn't expose this)

PR #380, opened 2015-07-24
fix Optimizely import for cocoapods 0.38.1

dcvz:
Due to a change in cocoapods the Optimizely framework is now built differently therefore Optimizely framework integration fails on build.

It won't pass any CI tests unless cocoapods 0.38.1 is being used.


f2prateek:
Should be fixed on master

PR #384, opened 2015-07-29
Kahuna iOS SDK 2.0.3 changes into Segment

sharekris:
NOTE : We are now using a new CocoaPod called "Kahuna". The old pod "KahunaSDK" is being deprecated. 

When this code gets released to CocoaPods, can someone from Segment change our pod in the Analytics.podspec from "KahunaSDK" to "Kahuna" ?

From :  "KahunaSDK": [ "1.0.570"] 
To :  "Kahuna" : ["2.0.3"]

That way the old POD is removed and replaced with the new POD. Here is our new POD spec https://github.com/CocoaPods/Specs/blob/master/Specs/Kahuna/2.0.3/Kahuna.podspec.json

### DIFF
--- a/Analytics.podspec
+++ b/Analytics.podspec
@@ -6,7 +6,7 @@ end
 
 Pod::Spec.new do |s|
   s.name            = "Analytics"
-  s.version         = "1.12.0"
+  s.version         = "1.12.1"
   s.summary         = "Segment analytics and marketing tools library for iOS."
   s.homepage        = "https://segment.com/libraries/ios"
   s.license         = { :type => "MIT", :file => "License.md" }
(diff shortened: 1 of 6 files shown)


f2prateek:
@sharekris You can do it by updating the reference here - https://github.com/segmentio/analytics-ios/blob/master/scripts/integrations.json#L65, and run `make build` which will update the references in the Podfile.

As for this PR, could you please squash these changes into one commit, and make sure that the CI passes? Let me know if you need any help with the CI errors.

sharekris:
@f2prateek I am referring to the Analytics.podspec itself change its reference from KahunaSDK to Kahuna since we now have a new pod called "Kahuna". https://github.com/CocoaPods/Specs/tree/master/Specs/Analytics

I am assuming customers integrating with Segment will use the 'Analytics' pod directly instead of including individual pods in their Podfile. 

I will create a new branch with the above changes squash committed and create a new PR.

f2prateek:
Ah I see, that will get updated when we make a release. We can't change it for the older releases, only newer releases will use the `Kahuna` pod.

PR #385, opened 2015-07-29
Squashed commit of the following:

sharekris:
Podfile.lock changed to use 2.0.3 after running make build.
    Merge pull request #1 from sharekris/updating_code_for_2.0.2
    Updating code for 2.0.3
    Removed tab.
    Changed to version 2.0.3 as that is the latest.
    Added call to setSDKWrapper method to send the wrapper name and version to our server.
    Modified the track event with quantity and revenue to trackEvent with count and value only when both revenue and quantity are specified in the properties.
    Removed commented out tests.
    Added OCMokito tests for Kahuna Integration. Made some code changes in SEGKahunaIntegration to allow OCMokito to mock the Kahuna object.
    Created SEGKahunaDefines.h header file and moved all static string definitions into that file.
    Modified getUserCredentials to use createUserCredentials. The latter is what we want to use since we want the request to identify detect non-overlaps. If we do a getUserCredential then it will always be over lapping credentials.
    Modified the Kahuna wrapper to use Kahuna instead of KahunaAnalytics SDK. Replaced the API setUserCredentialsWithKey with addCredential and then calling loginWithCredentials.

### DIFF
--- a/Analytics.xcodeproj/project.pbxproj
+++ b/Analytics.xcodeproj/project.pbxproj
@@ -23,12 +23,13 @@
 /* End PBXAggregateTarget section */
 
 /* Begin PBXBuildFile section */
+		019892FF1B67121A00ED875F /* SEGKahunaDefines.h in Headers */ = {isa = PBXBuildFile; fileRef = 019892FE1B67121A00ED875F /* SEGKahunaDefines.h */; };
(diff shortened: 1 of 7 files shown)


f2prateek:
Closed by #387

PR #393, opened 2015-07-31
Fix Kahuna push notification swizzling

f2prateek:
cc @sharekris @TorreyKahuna

### DIFF
--- a/Analytics/Integrations/Kahuna/SEGKahunaIntegration.h
+++ b/Analytics/Integrations/Kahuna/SEGKahunaIntegration.h
@@ -19,9 +19,10 @@
 
 
 @interface SEGKahunaPushMonitor : NSObject
-@property (nonatomic) NSDictionary *pushInfo;
-@property (nonatomic) UIApplicationState applicationState;
-@property (nonatomic) BOOL kahunaInitialized;
+@property (atomic) NSDictionary *pushInfo;
+@property (atomic) UIApplicationState applicationState;
+@property (atomic) BOOL kahunaInitialized;
+@property (atomic) NSError *failedToRegisterError;
 
 + (instancetype)sharedInstance;
 - (void)didFinishLaunching:(NSNotification *)userInfo;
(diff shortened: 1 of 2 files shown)


Analytics/Integrations/Kahuna/SEGKahunaIntegration.h
```diff
@@ -22,6 +22,7 @@
 @property (nonatomic) NSDictionary *pushInfo;
 @property (nonatomic) UIApplicationState applicationState;
 @property (nonatomic) BOOL kahunaInitialized;
```
f2prateek: @sharekris I noticed that you made these properties atomic in your change, is that strictly required?

sharekris: The property is being accessed in Start (which is in background thread) as well as App Delegate notifications (which is in Main thread). So made the properties atomic.

Analytics/Integrations/Kahuna/SEGKahunaIntegration.m
```diff
@@ -396,6 +396,7 @@ - (void)swizzleAppDelegateMethods
 - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotificationsWithError:(NSError *)error
 {
     @try {
+        [[SEGKahunaPushMonitor sharedInstance] failedToRegisterPush:error];
         if ([SEGKahunaPushMonitor sharedInstance].kahunaInitialized) {
             [KahunaAnalytics handleNotificationRegistrationFailure:error];
         }
```
sharekris: This "if ([SEGKahunaPushMonitor sharedInstance].kahunaInitialized) {" should not be present. The "failedToRegisterPush" takes care of the IF condition internally.

f2prateek: updated

Analytics/Integrations/Kahuna/SEGKahunaIntegration.h
f2prateek: Thanks, updated the properties to be atomic.

PR #396, opened 2015-08-01
Update Taplytics to 2.4.0

jonathannorris:
@f2prateek was still unable to run `make build` getting this when I remove the `Podfile.lock` file: 
```
Updating local specs repositories
Analyzing dependencies
[!] Unable to satisfy the following requirements:

- `MoEngage-iOS-SDK (= 1.4.3)` required by `Podfile`
- `MoEngage-iOS-SDK (= 1.4.3)` required by `Podfile`
make: *** [deps] Error 1
```

### DIFF
--- a/scripts/integrations.json
+++ b/scripts/integrations.json
@@ -109,7 +109,7 @@
       "name": "Taplytics",
       "dependencies": [{
         "name": "Taplytics",
-        "version": "2.3.12"
+        "version": "2.4.0"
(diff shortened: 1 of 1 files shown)


f2prateek:
@jonathannorris Thanks, I think the issue is from using CocoaPods 0.38.2 to build the library. Now that I've narrowed down the cause, we should be able to fix it.

PR #398, opened 2015-08-04
Kahuna SDK 2 changes for Segment Wrapper

sharekris:
Along with this merge, Analytics POD will have to be updated to use 'Kahuna' pod instead of 'KahunaSDK' pod.

### DIFF
--- a/Analytics.podspec
+++ b/Analytics.podspec
@@ -6,7 +6,7 @@ end
 
 Pod::Spec.new do |s|
   s.name            = "Analytics"
-  s.version         = "1.12.3"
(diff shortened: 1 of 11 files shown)


sharekris:
@f2prateek, The circleCI indicates that 1 test has failed, and it seems as if the "make test" command failed. When I run "make test" on the local machine it is able to perform all the "iOS Tests". When you get a chance can you look into it ?

travisjeffery:
Hey @sharekris thanks! Merged this is in, gonna look into what's up with circle

PR #403, opened 2015-08-12
Google Analytics: Add Support for Custom Dimensions and Metrics

sperand-io:
@f2prateek this obviously will need heavy review, especially since this integration does not yet have tests in place to see what I'm (probably) breaking.

Thought process so far:

 - set mapped *traits* as custom dimensions **on the tracker**, since they're inherently user or session scoped. like this example:

- set mapped *properties* of events and screen calls on their **respective hits**, since these are most typically hit-scoped. like this example:

Specifically, I'm not certain I'm correctly passing around / referencing / dereferencing the `GAIDictionaryBuilder*` objects (`hit` and `view`). I was going off this: https://github.com/edx/edx-app-ios/blob/master/Pods/GoogleAnalytics-iOS-SDK/GoogleAnalytics/Library/GAIDictionaryBuilder.h#L57-L69

### DIFF
--- a/Analytics/Integrations/GoogleAnalytics/SEGGoogleAnalyticsIntegration.m
+++ b/Analytics/Integrations/GoogleAnalytics/SEGGoogleAnalyticsIntegration.m
@@ -91,12 +91,14 @@
     if ([self shouldSendUserId])
         [[[GAI sharedInstance] defaultTracker] set:@"&uid" value:userId];
 
-    // We can set traits though. Iterate over a ll the traits and set them.
+    // We can set traits though. Iterate over all the traits and set them.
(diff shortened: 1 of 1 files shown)


sperand-io:
@f2prateek would you mind reviewing when you get a chance? Thanks!

f2prateek:
@sperand-io this is good to be shipped, I'm guessing we want the Android one published at the same time? https://github.com/segmentio/analytics-android/pull/333/files

sperand-io:
yup — i'll make the required edits over there by tonight

sperand-io:
@f2prateek fyi, just made a pass at the substring fix

PR #408, opened 2015-08-25
Updating Amplitude SDK v3.0.0

djih:
Updating Amplitude SDK to v3.0.0

* No longer log start/end sessions by default; however, users can re-enable them by setting trackSessionEvents to true. This is exposed in Segment via trackSessionEvents settings option.

(Also updated Cocoapods to 0.38?)

### DIFF
--- a/Analytics/Integrations/Amplitude/SEGAmplitudeIntegration.m
+++ b/Analytics/Integrations/Amplitude/SEGAmplitudeIntegration.m
@@ -29,6 +29,9 @@
 - (void)start
 {
     NSString *apiKey = [self.settings objectForKey:@"apiKey"];
+    if ([(NSNumber *)[self.settings objectForKey:@"trackSessionEvents"] boolValue]) {
+        [Amplitude instance].trackingSessionEvents = true;
+    }
     [[Amplitude instance] initializeApiKey:apiKey];
     SEGLog(@"AmplitudeIntegration initialized.");
     [super start];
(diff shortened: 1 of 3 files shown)


f2prateek:
There are some changes in 0.38.x that limit us from using it quite yet, so you'll have to resend the PR with 0.37.2 instead.

djih:
Reverted to Cocoapods 0.37.2. Please take another look

```
"trackSessionEvents": {
      "default": false,
      "description": "This will enable the sending of start and end session events to Amplitude, which are [no longer sent by default](https://github.com/amplitude/Amplitude-Android#tracking-sessions).",
      "label": "Track Session Events to Amplitude",
      "type": "boolean"
    }
```

djih:
@f2prateek do you actually want me to resubmit a PR or can you just re-review this one?

Analytics/Integrations/Amplitude/SEGAmplitudeIntegration.m
```diff
@@ -29,6 +29,9 @@ - (id)init
 - (void)start
 {
     NSString *apiKey = [self.settings objectForKey:@"apiKey"];
+    if ([(NSNumber *)[self.settings objectForKey:@"trackSessionEvents"] boolValue]) {
+        [Amplitude instance].trackingSessionEvents = true;
```
f2prateek: Is it ok to call this before the instance is initialized?

djih: Yes is it ok. In fact, the initialization can send a start session event, so that boolean needs to be set first.

https://github.com/amplitude/Amplitude-iOS#tracking-sessions

PR #419, opened 2015-09-14
Update Flurry and Localytics version

barrault01:
We experimenting crashes using the old version of Localytics, hope this fixe it.
Also update flurry sdk to support bitcode.


f2prateek:
There are some breaking API changes in the Flurry update, so upgrading won't be as simple as bumping the version.

barrault01:
can you at least update localytics?

f2prateek:
yup working on it, that not's not as straightforward either unfortunately.

PR #428, opened 2015-09-16
Re-adding UXCam - SDK version 2.3.2

richardgroves:
Latest UXCam SDK with updated cocoapod spec to get over previous issues


travisjeffery:
Thanks Richard! I ended up doing this in another commit and released it in 2.0.10.

PR #434, opened 2015-09-17
Support not setting traits on Mixpanel

f2prateek:
cc @wcjohnson11 @sperand-io 

Ref: https://segment.phacility.com/T265


f2prateek:
This is the current behvaiour:

First check if `setAllTraitsByDefault` is true.
If `setAllTraitsByDefault` is true, do the same thing as before and exit.

If `setAllTraitsByDefault` is false, check if there are any mixpanel options.
If there are no mixpanel options, then return early.

If there are mixpanel options, then iterate over all superProperties and peopleProperties in those options, and pull traits into 2 seperate dictionaries dependending on what's in those options.

Then map those dictionaries to Mixpanel's semantic properties.

PR #441, opened 2015-09-24
TvOS support

dbachrach:
Use TARGET_OS_TV to remove non-TvOS functionality.

### DIFF
--- a/Analytics/Helpers/SEGLocation.m
+++ b/Analytics/Helpers/SEGLocation.m
@@ -47,7 +47,9 @@
         self.geocoder = [[CLGeocoder alloc] init];
         self.locationManager = [[CLLocationManager alloc] init];
         self.locationManager.delegate = self;
+#if !(TARGET_OS_TV)
         [self.locationManager startUpdatingLocation];
+#endif
         self.syncQueue = dispatch_queue_create("io.segment.location.syncQueue", NULL);
     }
     return self;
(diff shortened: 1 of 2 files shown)


f2prateek:
@dbachrach have you been able to test this out on a tvOS device btw?

dbachrach:
No we didn't get lucky :disappointed: This is just on the simulator. Note, I also only did this for `Analytics/Segmentio`, and not all the other integrations.

f2prateek:
@dbachrach Thanks, yeah we don't have to worry about other integrations, since you can just do `pod Analytics/Segmentio` to use only use Segment.

orta:
I can confirm this works on an appleTV. I had to make podspec changes though; https://github.com/orta/analytics-ios/commit/7008b17f60735444538799d1b3fac893f7eedb0d

orta:
I intend in ARAnalytics to audit for tvOS support over time, for now I only added what I know works: https://github.com/orta/ARAnalytics/commit/1332c436ebc78e8bef5f3f8aa8d40c5ac985e963 - you may have to do the same if you're including static binaries

JasonGallagher:
The podfile probably needs to be updated to explicitly support tvOS too, right?

orta:
In this PR, yeah.

f2prateek:
Updated in #509, but not complete yet :(

PR #444, opened 2015-09-30
Bumped pod to 1.8.

gautamMoEngage:
Enabled modules.
Changed header path for MoEngage to match the pod settings.
Expecta, OCMock and OCHamcrest automatically updated via pod update.

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - Amplitude-iOS (3.0.0)
+  - Amplitude-iOS (3.0.1)
   - AppsFlyer-SDK (2.5.3.15.1)
   - Apptimize (2.11.0.1)
   - Bugsnag (4.0.7):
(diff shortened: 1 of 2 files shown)


gautamMoEngage:
@f2prateek - I'm getting the error "these headers need to be made public". Also, headers which are in Pods are not copied in the Segment folder. I see that pull requests (Apptimize, Amplitude, Taplytics) are also facing issues, while all they are trying to do it update the pod, same as me. I might be wrong, but can you look at it once if there might be an issue on your end.

f2prateek:
I was able to update all the other integrations, but not MoEngage https://cloudup.com/cOK5gUA7bKQ

gautamMoEngage:
Strange. I will have a look at it. I don't know why it's showing duplicate symbols.

gautamMoEngage:
Ok. So I found what is the issue. You are not supporting Bitcode, and you are probably using Xcode 6.x. My SDK is bitcode enabled and is built with Xcode 7. https://forums.developer.apple.com/thread/17529.

One of my clients needs Bitcode support. Can you try with Xcode 7 and building your SDK with bitcode enabled? If not, when do you plan to support bitcode?

francisco:
Hey @gautamMoEngage -

We don't yet have a specific timeline for supporting Bitcode. 

Would it be possible to resubmit the PR with your 1.7 release which according to our mutual client works if we don't support bitcode? Assuming doing so would also resolve the client's issue.

Francisco

PR #484, opened 2015-11-24
Proposed fix to allow the alias() call to succeede.

mcstoufer:
Both test suites passing with this fix in place. Not sure if test coverage actually handles this flow.


f2prateek:
Thanks! (you're right, tests had been removed to roll out v3 but we'll be adding them back over the next few weeks).

Issue #482, opened 2015-11-24
Custom Anonymous ID


alloy:
I’m wondering if there’s a reason for not allowing a custom Anonymous ID when using this SDK? [The API docs](https://segment.com/docs/spec/identify/#anonymous-id) make it seem like it’s perfectly fine to supply your own.

In our case we already have a UUID for unauthenticated users, which would make a lot of sense as the Anonymous ID.

mcstoufer:
As long as its unique, that would be ok. The initial call to alias() that to a real authenticated will work. However, later alias() calls from the same fixed anon UUID to a new real id will then have undefined behavior. At least that is what I gleaned from the docs as well.

alloy:
@mcstoufer I don’t really want to use the alias feature in this case, mainly because we use a lib that abstracts away details about the analytics provider (ARAnalytics).

f2prateek:
Adding a method to set the anonymous ID should be fine. We haven't had people ask us for this or provide a strong use case yet, so it hadn't been added yet.

PR #485, opened 2015-11-25
[Identify] Make Anonymous ID customisable.

alloy:
This is a first stab at making the Anonymous ID customisable, as asked about in #482. Some more background can be found in [this PR](https://github.com/artsy/eigen/pull/958) on our app.

This is very much a work in progress, I did the bare minimum to get _what we needed_ to work, so now I’m looking for feedback on:

* Is adding this feature conceptually ok?
* Would you prefer I refactor the `-getAnonymousId:` code into e.g. the getter of the `anonymousId` property, or what is the style you’d prefer?
* General code-style

### DIFF
--- a/Pod/Classes/SEGAnalytics.h
+++ b/Pod/Classes/SEGAnalytics.h
@@ -88,6 +88,8 @@
 
  @param traits        A dictionary of traits you know about the user. Things like: email, name, plan, etc.
 
+ @param options       A dictionary of options, such as the `@"anonymousId"` key. If no anonymous ID is specified one will be generated for you.
+
  @discussion
  When you learn more about who your user is, you can record that information with identify.
 
--- a/Pod/Classes/Integrations/SEGIdentifyPayload.h
+++ b/Pod/Classes/Integrations/SEGIdentifyPayload.h
@@ -6,9 +6,12 @@
 
 @property (nonatomic, readonly) NSString *userId;
 
+@property (nonatomic, readonly) NSString *anonymousId;
+
 @property (nonatomic, readonly) NSDictionary *traits;
 
 - (instancetype)initWithUserId:(NSString *)userId
+                   anonymousId:(NSString *)anonymousId
                         traits:(NSDictionary *)traits
                        context:(NSDictionary *)context
                   integrations:(NSDictionary *)integrations;
--- a/Pod/Classes/Integrations/SEGIdentifyPayload.m
+++ b/Pod/Classes/Integrations/SEGIdentifyPayload.m
@@ -4,12 +4,14 @@
 @implementation SEGIdentifyPayload
 
 - (instancetype)initWithUserId:(NSString *)userId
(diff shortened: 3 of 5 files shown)


f2prateek:
@alloy Thanks! This looks great, and should work as expected. I don't really have a preference for the style of the `-getAnonymousId:` method, unless you have a compelling reason to pick one or the other. General code style looks fine as well, I'll run it through our code formatter after merging.

My only question is if there is a better to expose this than hiding it in under the `options` parameter.
```objc
[[SEGAnalytics sharedAnalytics] identify:nil
                                           traits:@{ "b" : "c" }
                                          options:@{ @"anonymousId":@"d" }];

vs.

[[SEGAnalytics sharedAnalytics] setAnonymousID:@"d"];
[[SEGAnalytics sharedAnalytics] identify:nil traits:@{ "b" : "c" }];
```

alloy:
Nope, my only reason for asking it is purely to conform to your code base design ideas :)

Regarding exposing it, I’m fine with either, but maybe for consistency a method like 
`-[SEGAnalytics identify:anonymousID:traits:options:]` would be better? Whatever you choose, let me know and I’ll wrap this patch up.

mcstoufer:
It looks like any track() calls made before the setAnonymousId() and  identify() calls will miss using the intended custom anonymousId. Would it be up to the developer to implicitly understand this and they take the responsibility of setting the ID when they need it in the flow?

f2prateek:
@mcstoufer Yes, the current approach requires that if you want to use a custom anonymousID, you *must* call identify with the anonymous ID first (typically right after initializing the analytics client). Subsequent calls don't need to provide the anonymous ID.

alloy:
@mcstoufer Indeed. I’m guessing that that’s what most people already do anyways, or at least that it’s safe assumption.

f2prateek:
@alloy I think we can merge this change. Could you add a note in the [docs](https://github.com/segmentio/analytics-ios/blob/master/Pod/Classes/SEGAnalytics.h#L80-L92) to indicate how to use the `options` parameter to set a custom anonymous id?

Pod/Classes/SEGAnalytics.h
```diff
@@ -88,6 +88,8 @@
 
  @param traits        A dictionary of traits you know about the user. Things like: email, name, plan, etc.
 
+ @param options       A dictionary of options, such as the `@"anonymousId"` key. If no anonymous ID is specified one will be generated for you.
+
```
alloy: @f2prateek Is this good?

alloy:
Updated.

Pod/Classes/SEGAnalytics.h
f2prateek: Yup! We might want to clarify in the future that not bundled integrations will respect this, but we'll see if this confuses users first.

alloy: :+1:

PR #487, opened 2015-12-02
Update SEGSegmentIntegration.m to fix reachability

sk8rdave:
SCNetworkReachabilityGetFlags was always setting flags to 0
removing the protocol specifier caused it to work again


f2prateek:
Thanks!

Issue #440, opened 2015-09-24
TvOS Support


dbachrach:
When including the analytics Pod in a TvOS app, there are failures because of the use of certain CoreLocation features and CoreTelephony features.

Are there plans to update the SDK to support TvOS?

dbachrach:
I've pushed to my fork with the needed `#ifdef` fixes:

https://github.com/dbachrach/analytics-ios/commit/c8ef2b58af94d155ffabe49c412a89bcb4e52f1d

f2prateek:
Ah nice! This was definitely on our roadmap but haven't had time to look into it. Mind sending this as a PR? Seems pretty easy if that's all that was need, @travisjeffery and I can review it.

mootpointer:
This works, but there is a limitation with tvOS. You can't trust on the filesystem to be persistent. That means that things can disappear on you when the cache gets swept.

We've been using this fork for a little while, but have run into instances where users stopped being tracked for apparently no reason, then I remembered that Segment seems to store things in Application Support. That obviously is a Bad Thing™ when that's only a temporary directory.

charlie-gbox:
we really need the support. any estimated about it?

joeltello:
what's the status on this?

saverchenkov:
+1 for this one!

charlires:
We are all waiting for TvOS support :+1:

marcopetkovski:
+1 can't wait!

justindhill:
It looks like @mootpointer resolved the outstanding issues by augmenting the disk store with NSUserDefaults. Are you guys open to a PR that will add the appropriate ifdefs and add declaration of tvOS support in the Podspec?

f2prateek:
@justindhill sounds good, we had to revert the changes since we don't want to use NSUserDefaults on iOS (but do want it on tvOS). Happy to review a PR that falls back to NSUserDefaults on tvOS and updates the Podspec (bonus points if it supports Carthage!).

f2prateek:
Closed by #572 

To try it out:
`  pod 'Analytics', :git => 'https://github.com/segmentio/analytics-ios.git', :branch => 'dev'`

PR #488, opened 2015-12-02
Store `userId` and `anonymousId` in NSUserDefaults

mootpointer:
When looking for the `userId` and `anonymousId`, look first in
`NSUserDefaults`, then in the stored file. The stored file is persistent
in most cases, but on tvOS, the file system should be treated as
temporary. That means we lose the identity of the user is lost as soon
as the file system is cleared on the Apple TV.

This solution does not replace these file, but simply augments them,
using `NSUserDefaults` as a cache in front of the files. The only
rationale I can think of for using files is that sometimes apps blow
away their defaults.

There is probably another approach here which finds a happy medium, but
`NSUserDefaults` is all we've really got (unless you want to play with
the keychain, and I sure don't) on tvOS.

### DIFF
--- a/Pod/Classes/Internal/SEGSegmentIntegration.m
+++ b/Pod/Classes/Internal/SEGSegmentIntegration.m
@@ -22,6 +22,9 @@ NSString *const SEGSegmentRequestDidFailNotification = @"SegmentRequestDidFail";
 NSString *const SEGAdvertisingClassIdentifier = @"ASIdentifierManager";
 NSString *const SEGADClientClass = @"ADClient";
 
+NSString *const SEGUserIdKey = @"SEGUserId";
+NSString *const SEGAnonymousIdKey = @"SEGAnonymousId";
+
 static NSString *GenerateUUIDString()
 {
     CFUUIDRef theUUID = CFUUIDCreate(NULL);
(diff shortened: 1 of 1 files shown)


mootpointer:
This, or something similar will need to be implemented for #440 / #441 to work.

f2prateek:
Sorry for the delay, this looks good to me, but I don't think it's complete since if the FS is not persistent, we'd need to also do something similar for traits and the disk queue

mootpointer:
I ignored the disk queue because I thought we'd surely flush when we get backgrounded, which should cover the queue.

f2prateek:
The library flushes based on the queue size and a timer, but not with the app lifecycle. It's possible that an app might start and trigger events within the flush interval without saturating the disk queue size. These events would then be dropped.

Additionally, even if a flush is triggered, a network error (device is offline, Segment is down, general network flakiness) could prevent events from uploading. These events would also be dropped.

mootpointer:
The above commit adds similar `NSUserDefaults` persistence of the queue and traits.

Personally, at this point, I'd consider factoring the persistence out into another class, but I wanted to get it working for now.

Actually, if I were to refactor it, the queue itself would live in its own class as well, using the same persistence mechanism.

f2prateek:
Thanks!

PR #495, opened 2015-12-14
 Pass the userInfo dictionary as an argument to the app state handlers #495

Wenzhi:
This is critical for detecting push opens when the application does a cold-start (i.e. from suspended/closed) from a push click and the push-clicked signal (i.e. push payload in the launchOptions) is in the userInfo.

### DIFF
--- a/Pod/Classes/SEGAnalytics.m
+++ b/Pod/Classes/SEGAnalytics.m
@@ -158,18 +158,18 @@ NSString *SEGAnalyticsIntegrationDidStart = @"io.segment.analytics.integration.d
                             };
     });
     SEL selector = NSSelectorFromString(selectorMapping[note.name]);
-  if (selector) {
-    if ([note.name isEqualToString:UIApplicationDidFinishLaunchingNotification]) {
-      while (!self.initialized) {
-        [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow: 1.0]];
-      }
+    if (selector) {
+        if ([note.name isEqualToString:UIApplicationDidFinishLaunchingNotification]) {
+            while (!self.initialized) {
+                [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow: 1.0]];
+            }
+        }
+        NSArray *arguments = nil;
+        if (note != nil) {
+            arguments = @[note];
+        }
+        [self callIntegrationsWithSelector:selector arguments:arguments options:nil sync:true];
     }
-    NSArray *arguments = nil;
-    if (note != nil) {
-      arguments = @[note];
-    }
-    [self callIntegrationsWithSelector:selector arguments:arguments options:nil sync:true];
-  }
   
 }
 
--- a/Pod/Classes/SEGAnalytics.m
+++ b/Pod/Classes/SEGAnalytics.m
@@ -158,8 +158,12 @@ NSString *SEGAnalyticsIntegrationDidStart = @"io.segment.analytics.integration.d
                             };
     });
     SEL selector = NSSelectorFromString(selectorMapping[note.name]);
+    NSArray *arguments = nil;
+    if (note.userInfo != nil) {
+      arguments = @[note.userInfo];
+    }
     if (selector)
-        [self callIntegrationsWithSelector:selector arguments:nil options:nil sync:true];
+        [self callIntegrationsWithSelector:selector arguments:arguments options:nil sync:true];
 }
 
 #pragma mark - Public API
--- a/Pod/Classes/SEGAnalytics.m
+++ b/Pod/Classes/SEGAnalytics.m
@@ -158,12 +158,19 @@ NSString *SEGAnalyticsIntegrationDidStart = @"io.segment.analytics.integration.d
                             };
     });
     SEL selector = NSSelectorFromString(selectorMapping[note.name]);
-    NSArray *arguments = nil;
-    if (note.userInfo != nil) {
-      arguments = @[note.userInfo];
+  if (selector) {
+    if ([note.name isEqualToString:UIApplicationDidFinishLaunchingNotification]) {
+      while (!self.initialized) {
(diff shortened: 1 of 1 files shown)


f2prateek:
Pretty sure this doesn't actually work. Have you tested this?

Wenzhi:
@f2prateek I tested on iOS 9.2 on a 6s Plus, and the launch dictionary is passed down to `SEGAppboyIntegration`'s method `- (void)applicationDidFinishLaunching:(NSDictionary *)launchOptions`.

I can send you a sample project if you want.

f2prateek:
Interesting, I figured it wouldn't work because the method definition is different: https://github.com/segmentio/analytics-ios/blob/master/Pod/Classes/Integrations/SEGIntegration.h#L62. It doesn't accept an NSDictionary, but it accepts NSNotification.

f2prateek:
@Wenzhi could you explain  how the `- (void)applicationDidFinishLaunching:(NSDictionary *)launchOptions` was being invoked even though the method signature is different https://github.com/segmentio/analytics-ios/blob/master/Pod/Classes/Integrations/SEGIntegration.h#L62

Wenzhi:
Hey @f2prateek,

I originally made the method accept an NSDictionary by following the scheme of your callbacks for remote notifications (it worked because objective C will call that method if it exists regardless of what the protocol says).  I’ve updated the pull to correct the method signature.  I’ve also updated to keep didFinishLaunching running until the SDKs finish launching. By doing so, we can collect the push statistics and handle the HTTP-links/deep-links in the push payload.  A lot of push detection and handling depends on application state, and without this, the signal gets lost.

f2prateek:
I have to follow up questions:

1. We're already calling this method synchronously on the integration so you should get the correct application state — why do we need to keep the `didFinishLaunching` method running in addition to this?

2. What application state are you recording? Can you share a concrete example so I can understand better?

Wenzhi:
Hey @f2prateek,

The issue we’re seeing is that the applicationDidFinishLaunching callback seems to not be getting called synchronously - it’s happening before initialized is set to true, so it’s doing it async.  We need to look at application state because the push signal in applicationDidFinishLaunching (i.e. a push payload) looks the same when the app is launched in the background and when the app is being launched into the foreground.  We need the app state to determine if it’s a push open or a background delivery.(edited)

Furthermore, we route behavior through shared push handling code, which relies on similar application state signal (in relation to didReceiveRemoteNotification calls).  If didFinishLaunching is called from an Active state, it looks like a push received in the foreground (which we log differently than a regular push because the system doesn’t show it)

f2prateek:
I think we might have to chalk this down to a limitation as a segment integration. We'll try to deliver it synchronously when we can, but obviously it's not always possible, and I'd rather not block the lifecycle methods for this.

Wenzhi:
How about only keep the method running when Appboy is intergrated?
```
if ([note.name isEqualToString:UIApplicationDidFinishLaunchingNotification] && 
    NSClassFromString(@"SEGAppboyIntegrationFactory") != nil) {
+      while (!self.initialized) {
+        [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow: 1.0]];
+      }
+    }
```

f2prateek:
That wouldn't be better since then we'd have integration specific code in the core library.

The main problem here is the 1s hardcoded duration. What guarantees that Appboy (and any other integration needs this) gets the event within a second and processes it? it could easily be outside this window and we'd back to square one.

Wenzhi:
Hey Prateek, that code is in a while loop, so it'll keep waiting till `self.initialized` is YES (checking in 1 second intervals).  I think that should be a fairly clean and efficient way to wait for everything to be set up before calling the integrations - please let me know what you think.

f2prateek:
But in that case, what happens if the integrations never get initialized? This scenario is not unlikely if the user's device is offline on the very first time they launch the app — we never get any settings and don't have anything cached so we're blocking continuously and indefinitely.

Pod/Classes/SEGAnalytics.m
```diff
@@ -158,8 +158,19 @@ - (void)handleAppStateNotification:(NSNotification *)note
                             };
     });
     SEL selector = NSSelectorFromString(selectorMapping[note.name]);
-    if (selector)
-        [self callIntegrationsWithSelector:selector arguments:nil options:nil sync:true];
+  if (selector) {
```
f2prateek: Indentation here is slightly off (I can fix this later as well).

Pod/Classes/SEGAnalytics.m
```diff
@@ -158,8 +158,19 @@ - (void)handleAppStateNotification:(NSNotification *)note
...
+    if ([note.name isEqualToString:UIApplicationDidFinishLaunchingNotification]) {
+      while (!self.initialized) {
+        [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode beforeDate:[NSDate dateWithTimeIntervalSinceNow: 1.0]];
+      }
+    }
+    NSArray *arguments = nil;
+    if (note != nil) {
+      arguments = @[note];
```
f2prateek: Should we update the definitions here to indicate that a dictionary might be passed in as an argument? https://github.com/segmentio/analytics-ios/blob/master/Pod/Classes/Integrations/SEGIntegration.h#L62-L67

f2prateek:
Last 2 questions, lgtm to otherwise!

Pod/Classes/SEGAnalytics.m
Wenzhi: I fixed it. Thank you for catching it.

Pod/Classes/SEGAnalytics.m
Wenzhi: I think those definitions are fine. I updated my code so a notification is passed in instead of a dictionary.

f2prateek: Hey @Wenzhi, I think the code is still the same. Mind updating that and we can merge this and make a release?

f2prateek: Ah nvm I see what you meant.

f2prateek:
Sorry @Wenzhi I wanted to make a release today so I undid the PR for now, could you re-open it?

PR #502, opened 2016-01-08
Fix Crash: Mutable queue and traits

enthouan:
Queue and Traits properties should return a mutable object. If they are retrieved from NSUserDefaults they are immutable by default, therefore mutableCopy is needed.

I had to downgrade segment pod to the previous version because of this. It was randomly crashing at launch with AppBoy integration.


f2prateek:
Thanks!

PR #503, opened 2016-01-08
Fix location updates issues

aleite:
This PR should address several issues found on version 3.0.4:

- When SEGLocation is initialized from a background thread, it never receives location updates, for this reason I am initializing the locationManager in the main thread.
- SEGSegmentIntegration was initializing SEGLocation every time liveContext method was called (which happens pretty frequently) and shouldUseLocationServices = true. By initializing it all the time we were deallocating the previous instance and allocating a new one and starting the location manager again. This was causing a crash when reverseGeocodeLocation completion handler was called and the syncQueue had already being deallocated. To fix this crash I updated the code to initialize SEGLocation only once.
- SEGLocation->didUpdateLocations was handling location updates without taking into consideration the last time the location was updated. I added code to take location timestamp into consideration and defined LOCATION_AGE (5 min), so any location older than 5 min is automatically discarded.
- There were no calls to stopUpdatingLocation, so once we called startUpdatingLocation we would keep receiving location updates all the time and we would try to reverse geocode all these locations. I noticed that calls to reverse geocode where being made at an approximate rate of 1call/second. In summary this would keep the GPS on and also use network resources to do the reverse geocode. To fix this problem I updated the code to stop getting updates after we get a successful reverse geocode response and we will only process the next location updates after LOCATION_AGE (5 min).
- SEGSegmentIntegration->queue, SEGSegmentIntegration->traits were returning instances of NSArray and NSDictionary instead of their mutable versions. I updated the code to make sure the proper mutable versions are returned. This was causing exceptions while trying to queue a new payload (queuePayload).

### DIFF
--- a/Pod/Classes/Internal/SEGLocation.h
+++ b/Pod/Classes/Internal/SEGLocation.h
@@ -23,4 +23,6 @@
 @property (nonatomic, copy, readonly) NSDictionary *addressDictionary;
 @property (nonatomic, assign, readonly) BOOL hasKnownLocation;
 
+- (void)startUpdatingLocation;
+
 @end
--- a/Pod/Classes/Internal/SEGLocation.m
+++ b/Pod/Classes/Internal/SEGLocation.m
@@ -26,6 +26,7 @@
         return result;                                                                             \
     }
 
+#define LOCATION_AGE 300.0 // 5 minutes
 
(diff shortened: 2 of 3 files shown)


f2prateek:
Thanks! Please send these as different PRs each time, we'd already received a fix for the last one.

PR #505, opened 2016-01-11
Fix crash in SEGAnalytics, func invocationForSelector

Maxim-Inv:
Crashed in SEGAnalytics, func invocationForSelector, [invocation setArgument:&argument atIndex:i + 2], because applicationDidEnterBackground in SEGSegmentIntegration has no one parameter.


f2prateek:
Fixed in 3.0.6, will be updating the method definition so integrations can take advantage of push notification data.

PR #510, opened 2016-02-01
Added missing initializer in SEGAnalytics.h & fixed return types

fluidsonic:
The initializer `-initWithConfiguration:` is missing in the public header file and thus unusable.
It is referred to by the deprecation message of `-initWithWriteKey:` and must be used when creating a non-singleton instance of `SEGAnalytics`.

This pull request adds it to the public header file.

Also this pull request fixes wrong return types specified for the initializers. 
They now return `instancetype` instead of `id`.

### DIFF
--- a/Pod/Classes/SEGAnalytics.h
+++ b/Pod/Classes/SEGAnalytics.h
@@ -56,6 +56,13 @@
 @property (nonatomic, strong, readonly) SEGAnalyticsConfiguration *configuration;
 
 /**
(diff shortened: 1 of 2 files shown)


f2prateek:
Thanks!
FYI you can't really use Analytics iOS other than as a singleton (either as provided or creating one instance and sharing it everywhere yourself) because the persistent caches are not namespaced.

PR #519, opened 2016-03-19
Fixed crashes due to NSNotificationCenter observers not being removed

fluidsonic:
When an `SEGAnalytics` instance gets destroyed the app will crash on certain notifications sent by `NSNotificationCenter` because `SEGAnalytics` never removes itself as an observer.

This pull request properly removes the observers.


f2prateek:
Thanks!

PR #524, opened 2016-05-16
Fix notification token setting in the segment integration

cfraz89:
The segment integration is currently using the old-school selector of  registerForRemoteNotificationsWithDeviceToken:options to handle recieving a notification token. Therefore when [SEGAnalytics registeredForRemoteNotificationsWithDeviceToken:] is called, the integration does not receive the forwarded call.

This PR updates the selector to registeredForRemoteNotificationsWithDeviceToken: so that it once again receives the forwarded calls.


f2prateek:
Thanks

PR #525, opened 2016-05-22
Automatic Screen Tracking

f2prateek:


Pod/Classes/SEGAnalytics.m
```diff
@@ -5,11 +5,11 @@
 #import "SEGAnalyticsUtils.h"
 #import "SEGAnalyticsRequest.h"
 #import "SEGAnalytics.h"
-
 #import "SEGIntegrationFactory.h"
 #import "SEGIntegration.h"
-#import <objc/runtime.h>
 #import "SEGSegmentIntegrationFactory.h"
+#import "UIViewController+SEGScreen.h"
+#import <objc/runtime.h>
```
tonyxiao: Unneeded import?

Pod/Classes/Internal/UIViewController+SEGScreen.m
```diff
@@ -0,0 +1,76 @@
...
+                                method_getImplementation(originalMethod),
+                                method_getTypeEncoding(originalMethod));
+        } else {
+            method_exchangeImplementations(originalMethod, swizzledMethod);
+        }
+    });
+}
+
+
++ (UIViewController *)topViewController
```
tonyxiao: We should prefix our method names if we were to add category to global classes such as `UIViewController`. something like `seg_topViewController`

Pod/Classes/Internal/UIViewController+SEGScreen.m
```diff
@@ -0,0 +1,76 @@
...
+        UINavigationController *navigationController = (UINavigationController *)rootViewController.presentedViewController;
+        UIViewController *lastViewController = [[navigationController viewControllers] lastObject];
+        return [self topViewController:lastViewController];
+    }
+
+    UIViewController *presentedViewController = (UIViewController *)rootViewController.presentedViewController;
+    return [self topViewController:presentedViewController];
+}
+
+- (void)screen_viewDidAppear:(BOOL)animated
```
tonyxiao: I think should probably use the same consistent prefix as well - `seg_`.

Pod/Classes/Internal/UIViewController+SEGScreen.m
f2prateek: Ah good point

Pod/Classes/Internal/UIViewController+SEGScreen.m
f2prateek: Gotcha

Pod/Classes/SEGAnalytics.m
f2prateek: Good catch! This was actually in the file from earlier for some reason....

f2prateek: Ah we use it in forwarding events to methods e.g. https://github.com/segmentio/analytics-ios/blob/master/Pod/Classes/SEGAnalytics.m#L553

tonyxiao:
👍 

PS: `xcodeproj` files are so hard to diff.

f2prateek:
@tonyxiao Haha agree, is it safe to simply uncheck them Git? I copied the .gitignore config from somewhere and it didn't have it.

Maybe we could check it in but treat it as a binary file http://shanesbrain.net/2008/7/9/using-xcode-with-git

PR #528, opened 2016-05-31
revert NSUserDefaults Queue and Traits

wcjohnson11:
@tonyxiao 

This PR reverts [this change](https://github.com/segmentio/analytics-ios/commit/3aa84ccdb7e69623981dc57db1ec92283a80a260) from back in December that stored the queue in `NSUserDefaults` instead of writing to disk. I've tested this locally and on my device and confirmed that cached traits and events work as expected.

We'll probably want to do a more in-depth refactor of how we manage the queue but reverting to this method should be a good stopgap measure.

### DIFF
--- a/Pod/Classes/Internal/SEGSegmentIntegration.m
+++ b/Pod/Classes/Internal/SEGSegmentIntegration.m
@@ -24,8 +24,6 @@ NSString *const SEGADClientClass = @"ADClient";
 
 NSString *const SEGUserIdKey = @"SEGUserId";
 NSString *const SEGAnonymousIdKey = @"SEGAnonymousId";
-NSString *const SEGQueueKey = @"SEGQueue";
-NSString *const SEGTraitsKey = @"SEGTraits";
 
 static NSString *GenerateUUIDString()
 {
@@ -254,7 +252,6 @@ static BOOL GetAdTrackingEnabled()
 {
     [self dispatchBackground:^{
         self.userId = userId;
-        [[NSUserDefaults standardUserDefaults] setValue:userId forKey:SEGUserIdKey];
         [self.userId writeToURL:self.userIDURL atomically:YES encoding:NSUTF8StringEncoding error:NULL];
     }];
 }
@@ -272,7 +269,6 @@ static BOOL GetAdTrackingEnabled()
 {
     [self dispatchBackground:^{
         [self.traits addEntriesFromDictionary:traits];
-        [[NSUserDefaults standardUserDefaults] setObject:[self.traits copy] forKey:SEGTraitsKey];
         [[self.traits copy] writeToURL:self.traitsURL atomically:YES];
     }];
 }
@@ -394,7 +390,7 @@ static BOOL GetAdTrackingEnabled()
 {
     @try {
         [self.queue addObject:payload];
-        [self persistQueue];
+        [[self.queue copy] writeToURL:[self queueURL] atomically:YES];
         [self flushQueueByLength];
 
     }
@@ -466,8 +462,6 @@ static BOOL GetAdTrackingEnabled()
     [self dispatchBackgroundAndWait:^{
         [[NSUserDefaults standardUserDefaults] setValue:nil forKey:SEGUserIdKey];
         [[NSUserDefaults standardUserDefaults] setValue:nil forKey:SEGAnonymousIdKey];
-        [[NSUserDefaults standardUserDefaults] setValue:@[] forKey:SEGQueueKey];
-        [[NSUserDefaults standardUserDefaults] setValue:nil forKey:SEGTraitsKey];
         [[NSFileManager defaultManager] removeItemAtURL:self.userIDURL error:NULL];
         [[NSFileManager defaultManager] removeItemAtURL:self.traitsURL error:NULL];
         [[NSFileManager defaultManager] removeItemAtURL:self.queueURL error:NULL];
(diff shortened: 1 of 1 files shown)


tonyxiao:
@wcjohnson11 What is this PR trying to accomplish? Was saving to NSUserDefault causing problems?

f2prateek:
LGTM

Pod/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -530,25 +528,35 @@ - (void)applicationWillTerminate
 {
     [self dispatchBackgroundAndWait:^{
         if (self.queue.count)
-
-            [self persistQueue];
+            [[self.queue copy] writeToURL:self.queueURL atomically:YES];
     }];
 }
 
+-(void)applicationDidFinishLaunching
```
wcjohnson11: I added this application state change event so we would check if there is a previous version of our SDK and then we would check if there's an existing queue/traits in NSUserDefaults and if it exists, then copy it to disk and delete the copy from NSUserDefaults. I'm not sure this will work and thinking there's a better way to do this so let me know what you think @tonyxiao @f2prateek

wcjohnson11:
After chatting with @tonyxiao and @sperand-io, we decided that we would provide some backwards compatibility that would check for a queue/traits in NSUserDefaults and write them to disk when the app is installed. I modeled it loosely off of [the lifecycle events code](https://github.com/segmentio/analytics-ios/blob/359c36b0d607f4ca496e010aa931ac5a87730b5a/Pod/Classes/SEGAnalytics.m#L147-L181) but am thinking there's a better way to handle this.

Pod/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -528,6 +532,17 @@ - (void)applicationWillTerminate
...
+{
+    [self beginBackgroundTask];
+    // We will check if there is a previous build of the app where traits/queue was stored in NSQ
```
tonyxiao: Too much `NSQ` fun? :D You mean `NSUserDefaults`? :)

Pod/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -528,6 +532,17 @@ - (void)applicationWillTerminate
     }];
 }
 
+-(void)applicationDidFinishLaunching
+{
+    [self beginBackgroundTask];
```
tonyxiao: Shouldn't migration be the very first thing that happens after application startup?

Pod/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -528,6 +532,17 @@ - (void)applicationWillTerminate
     }];
 }
 
+-(void)applicationDidFinishLaunching
```
tonyxiao: I think it's more appropriate for migration to happen at `SEGSegmentIntegration` initialization time rather than waiting until `applicationDidFinishLaunching`. The important bit here is to ensure that old events are migrated prior to the rest of `SEGSegmentIntegration` functionality happening.

Pod/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -528,6 +532,17 @@ - (void)applicationWillTerminate
...
+{
+    [self beginBackgroundTask];
+    // We will check if there is a previous build of the app where traits/queue was stored in NSQ
+    // If there is then we will store the existing queue and traits in the new location
+    if ([self hasPreviousVersion]) {
```
tonyxiao: Why do we need to check for version explicitly? Can we just check the existence of `SEGQueue` and `SEGTraits` key directly in `NSUserDefaults` and if they exist then migrate them?

Pod/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -592,4 +607,52 @@ - (NSString *)getUserId
...
+
+- (void)transferTraits
+{
+    NSDictionary *oldTraits = [self NSUserDefaultTraits];
+    // If there is a traits dict in NSUserDefaults
+    if (oldTraits) {
+        // Remove the old traits
+        [[NSUserDefaults standardUserDefaults] removeObjectForKey:SEGTraitsKey];
+        SEGLog(@"[[NSUserDefaults standardUserDefaults] removeObjectForKey: %@]", SEGTraitsKey);
+        [[oldTraits copy] writeToURL:[self traitsURL] atomically:YES];
```
tonyxiao: Could we possibly be over-writing existing data there? Is it possible for previous version of the SDK to somehow have both stored data in `NSUserDefaults` as well as on disk as files?

Pod/Classes/Internal/SEGSegmentIntegration.m
tonyxiao: Also, we *might* want to dispatch this to a background (serial) queue and call it. Reading / writing large files can take up quite a bit of time and we might be blockign the application from starting / UI from showing if we perform too much work here. (e.g. reading / writing 40mb files)

tonyxiao: Although if that was async we'd have to be careful how it gets loaded from disk in the accessor so we do not run into race conditions.

tonyxiao:
@wcjohnson11 thoughts added

Pod/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -93,6 +95,13 @@ - (id)initWithAnalytics:(SEGAnalytics *)analytics
         self.serialQueue = seg_dispatch_queue_create_specific("io.segment.analytics.segmentio", DISPATCH_QUEUE_SERIAL);
         self.flushTaskID = UIBackgroundTaskInvalid;
         self.analytics = analytics;
+        // Check for previous queue/track data in NSUserDefaults and store to disk
```
wcjohnson11: Would dispatching to a background queue look like wrapping lines 99 ->104 or would that be inside the body of the transferQueue and transferTraits methods? @tonyxiao

wcjohnson11:
@tonyxiao I made some changes, moving the logic for this change into the initialization and checking for the array and traits in NSUserDefaults instead of a previous version. I didn't implement dispatch to a background serial queue, but left notes inline there. Let me know what you think.

PR #535, opened 2016-06-03
remove transferQueue and transferTraits methods

wcjohnson11:
This PR removes transferring the queue and traits to disk since they'd previously been written to both locations. These methods could have duplicated events in users queues. It now only checks for the queue and traits in NSUserDefaults and then removes them if present. @tonyxiao @f2prateek


f2prateek:
LGTM.

PR #538, opened 2016-06-06
Added static CTTelephonyNetworkInfo as fix for iOS Bug

johnryan:
Created a static instance of CTTelephonyNetworkInfo to avoid a crash caused by a bug in iOS see:
http://stackoverflow.com/questions/14238586/coretelephony-crash

### DIFF
--- a/Pod/Classes/Internal/SEGSegmentIntegration.m
+++ b/Pod/Classes/Internal/SEGSegmentIntegration.m
@@ -106,6 +106,15 @@ static BOOL GetAdTrackingEnabled()
     return self;
 }
 
+/*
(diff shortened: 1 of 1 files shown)


f2prateek:
Thanks! LGTM, could you add a comment above the static declaration indicating why we do this? Something like:

```
/*
* There is an iOS bug that causes instances of the CTTelephonyNetworkInfo class to 
* sometimes get notifications after they have been deallocated.
* Instead of instantiating, using, and releasing instances you * must instead retain 
* and never release them to work around the bug.
* 
* Ref: http://stackoverflow.com/questions/14238586/coretelephony-crash
*/
```

johnryan:
Yep, just added that in, Thanks.

f2prateek:
Thanks!

PR #542, opened 2016-06-07
Handle root being a navigation controller

voidref:
The automatic screen event code did not handle the case where the root view controller was a UINavigationController.

The code would often just report the top level Nav as it would not always have a presented VC.

This proposed change checks for a top level nav controller, and steps into either it's presented controller or it's last in the list of view controllers as appropriate.

### DIFF
--- a/Pod/Classes/Internal/UIViewController+SEGScreen.m
+++ b/Pod/Classes/Internal/UIViewController+SEGScreen.m
@@ -38,37 +38,31 @@
 + (UIViewController *)seg_topViewController
 {
     UIViewController *root = [UIApplication sharedApplication].delegate.window.rootViewController;
-    
-    if ([root isKindOfClass:[UINavigationController class]]) {
-        UIViewController* presented = root.presentedViewController;
-        
-        if (nil == presented) {
-            presented = ((UINavigationController *)root).viewControllers.lastObject;
-        }
-        
-        if (nil != presented) {
-            root = presented;
-        }
-    }
-    
     return [self seg_topViewController:root];
 }
 
 + (UIViewController *)seg_topViewController:(UIViewController *)rootViewController
 {
+    if ([rootViewController isKindOfClass:[UINavigationController class]]) {
+        UIViewController* presented = rootViewController.presentedViewController;
+        
+        if (nil == presented) {
+            presented = ((UINavigationController *)rootViewController).viewControllers.lastObject;
+        }
+        
+        if (nil != presented) {
+            rootViewController = presented;
+        }
+    }
+    
     if (rootViewController.presentedViewController == nil) {
         return rootViewController;
-    }
-
-    if ([rootViewController.presentedViewController isKindOfClass:[UINavigationController class]]) {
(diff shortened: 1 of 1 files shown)


voidref:
It occurs to me that the root logic may actually need to be moved into the recursive function for navigation controller detections.

f2prateek:
@voidref I was about to suggest that :) Do you think that's a good idea as well?

voidref:
Yeah, it really should be moved into the recursive function, people nest view controllers all the time in more complex apps (even though they might ought not to.) =)

f2prateek:
Cool let me know when it's been updated and I can take a 2nd look!

voidref:
OK, updated with the latest logic for ya! Did a quick test on my app, and it doesn't work 'worse', but I don't have a comprehensive test suite for this (you might want to put something together).

f2prateek:
I think we simplify it to:

```
+ (UIViewController *)seg_topViewController:(UIViewController *)rootViewController
{
    if ([rootViewController isKindOfClass:[UINavigationController class]]) {
        UIViewController *lastViewController = [[(UINavigationController *)rootViewController viewControllers] lastObject];
        return [self seg_topViewController:lastViewController];
    }

    UIViewController *presentedViewController = (UIViewController *)rootViewController.presentedViewController;
     if (presentedViewController == nil) {
        return rootViewController;
    }
    return [self seg_topViewController:presentedViewController];
}
```

voidref:
I am afraid that doesn't quite work, because you might have a modal view presented on top of the navigation controller, it would then be the 'presentedViewController', and not one of the ones in the viewControllers list (which are the pushed view controllers)

f2prateek:
Ah I see, so this is more appropriate:

```
+ (UIViewController *)seg_topViewController:(UIViewController *)rootViewController
{
    if ([rootViewController isKindOfClass:[UINavigationController class]]) {
        UIViewController *presentedViewController = (UIViewController *)rootViewController.presentedViewController;
        if (presentedViewController == nil) {
            return [self seg_topViewController:presentedViewController];
        }
        
        UIViewController *lastViewController = [[(UINavigationController *)rootViewController viewControllers] lastObject];
        return [self seg_topViewController:lastViewController];
    }

    UIViewController *presentedViewController = (UIViewController *)rootViewController.presentedViewController;
    if (presentedViewController == nil) {
        return rootViewController;
    }
    return [self seg_topViewController:presentedViewController];
}
```

voidref:
Ah, no, not quite, if you have a nil `presentedViewController` on a nav controller, you need to use the last viewController from the array.

That impl will return nil, it looks like.

f2prateek:
Oops I meant that for that first check to be `        if (presentedViewController != nil) {`

voidref:
This works except for a extremely rare corner case where the nav controller has no subviews or presented views, but I can't think of a time that wouldn't just be an application error. If you are ok with returning nil in this case, you are good to go.

f2prateek:
Yup I think for now returning nil will work. This is a fairly new feature as well so it might not work in every case — so I'm hoping if this limitation is a blocker for anyone in practice, we can improve it over time.

f2prateek:
Closed by #547

PR #555, opened 2016-06-22
Fix for when a title may come back as empty string

shaun-seek:
There are scenarios that an alert view controller for instance could be displayed without specifying a title.  In that instance the screenTitle.length > 0 assertion would fail.

This fix simply checks that name.length == 0 and if it is then try and generate a useful name.


f2prateek:
SGTM, can you send this against the `dev` branch instead?

shaun-seek:
Will do!

PR #560, opened 2016-06-23
Initialize flush timer on the main thread.

### LINKED ISSUE #558: Flush Timer not fired in SEGSegmentIntegration.m
https://github.com/segmentio/analytics-ios/blob/master/Pod/Classes/Internal/SEGSegmentIntegration.m#L93

This timer is being created in the io.segment.analytics thread which doesn't have run loops so the timer never fires. 

You can either create the timer on the main thread, or add it too the mainRunLoop

http://stackoverflow.com/questions/12323531/difference-in-scheduling-nstimer-in-main-thread-and-background-thread

f2prateek:
closes #558


underscoretang:
thanks!

PR #563, opened 2016-06-25
Replace Foundation import with UIKit import

indragiek:
We were integrating this library using GYP instead of CocoaPods and there was a compilation issue because `UIViewController` is defined in `UIKit` rather than `Foundation`.


f2prateek:
Thanks!

PR #568, opened 2016-06-30
correctly skip forwarding disabled events in the tracking plan

f2prateek:
Previously events that were marked as disabled in the tracking plan
would still be sent to Segment (and other integrations, both client side
and server side).

The bug was that we were accidentally checking the class name of the
SEGTrackPayload instead of the `event` in the payload.


f2prateek:
Travis CI is experiencing downtime so merging this having been tested locally.

PR #570, opened 2016-06-30
tvOS support

justindhill:
This PR addresses #440. It fixes compilation issues under tvOS as well as re-adding NSUserDefaults storage of the trait collection and event queue under tvOS. We check for the file on disk as normal, then fall back onto the NSUserDefaults entry if it doesn't exist. In addition, the Podspec was updated with a version bump to 3.2.6 and a new deployment target for tvOS (9.0).

Couple of things:

1. I think it would actually be better to simply store the event queue and traits entirely in NSUserDefaults on tvOS and eschew the file representation altogether. I didn't do this because @f2prateek expressed a desire to only fall back on NSUserDefaults in #440.

2. I've never used Carthage, but I looked into it and it appears that you must have a separate target in your Xcode project for each device target, as it doesn't do any project generation. This presents an issue with the current project structure - analytics-ios uses the development pod target created by CocoaPods and has no explicit target, so I would have had to significantly alter the structure, and I wasn't feeling particularly presumptuous, so I left it alone. :)


justindhill:
It looks like the Travis build failed because the Xcode version selected for the project is too old - it doesn't contain the tvOS SDK. If you bump the version to Xcode 7.3, it should build fine.

Analytics.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "Analytics"
-  s.version          = "3.2.5"
+  s.version          = "3.2.6"
```
f2prateek: Can you leave this as is? We'll bump it as part of the release process.

justindhill: 👍

f2prateek:
@justindhill Thanks! Some comments:

1. Feel free to bump the Xcode version in the travis config.
2. Sorry for the confusion - I do agree that we should skip the file persistence on tvOS altogether, and use NSUserDefaults instead. On iOS - we can leave the current behaviour as is.

justindhill:
After I bumped the Xcode version in Travis it looks like we timed out on `[Info] Collecting info for testables...`, but it passed on my machine!

image

f2prateek:
@justindhill no worries, I'll test this locally as well and we can skip waiting on the CI to pass (it's been a bit flaky recently).

Could you reopen this PR against `dev` instead of `master` as well — https://github.com/segmentio/analytics-ios/blob/dev/CONTRIBUTING.md

justindhill:
👍 re-opened as #571 into `dev`

PR #509, opened 2016-02-01
TvOS support

f2prateek:
Just a rebase + some necessary Podspec updates.

Closes #441

PR #571, opened 2016-07-01
tvOS support

justindhill:


f2prateek:
Hey @justindhill - looks like the compilation works fine but the Podspec does not validate. you can run `make lint` locally to see the output yourself.

https://cloudup.com/cfQaMXWRBEO

```
pod lib lint

 -> Analytics (3.2.5)
    - WARN  | [tvOS] xcodebuild:  /Users/prateek/dev/src/github.com/segmentio/analytics-ios/Analytics/Classes/Internal/SEGAnalyticsRequest.m:34:48: warning: 'initWithRequest:delegate:startImmediately:' is deprecated: first deprecated in tvOS 9.0 - Use NSURLSession (see NSURLSession.h) [-Wdeprecated-declarations]
    - NOTE  | [tvOS] xcodebuild:  /Applications/Xcode.app/Contents/Developer/Platforms/AppleTVSimulator.platform/Developer/SDKs/AppleTVSimulator9.2.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSURLConnection.h:117:1: note: 'initWithRequest:delegate:startImmediately:' has been explicitly marked deprecated here
    - WARN  | [tvOS] xcodebuild:  /Users/prateek/dev/src/github.com/segmentio/analytics-ios/Analytics/Classes/Internal/SEGLocation.m:38:17: warning: method definition for 'startUpdatingLocation' not found [-Wincomplete-implementation]
    - NOTE  | [tvOS] xcodebuild:  /Users/prateek/dev/src/github.com/segmentio/analytics-ios/Analytics/Classes/Internal/SEGLocation.h:18:1: note: method 'startUpdatingLocation' declared here
    - WARN  | [tvOS] xcodebuild:  /Users/prateek/dev/src/github.com/segmentio/analytics-ios/Analytics/Classes/Internal/SEGSegmentIntegration.m:649:12: warning: unused variable 'url' [-Wunused-variable]

[!] Analytics did not pass validation, due to 3 warnings (but you can use `--allow-warnings` to ignore them).
You can use the `--no-clean` option to inspect any issue.
make: *** [lint] Error 1
```

Some of these are actually the same errors I was stuck at last time I looked into this https://github.com/segmentio/analytics-ios/pull/509#issuecomment-177738562. The good thing is there are no errors now, just warnings!

So if we can fix the warnings that'd be great. Additionaly I'd like to do some manual testing on tvOS for a sanity check as well.

justindhill:
Hey @f2prateek, resolved all but one of the linting errors. The last one is caused by the use of NSURLConnection in `SEGSegmentIntegration`. You'll need to switch to using NSURLSession to resolve that one.

justindhill:
However, that one is a warning and can be resolved at your leisure.

f2prateek:
@justindhill Thanks, we can take care of that separately- we do have an open issue for migrating to NSUrlSession but haven't had time to tackle it. 

Do you know if it's possible to have CocoaPods ignore a specific error (other than supressing all with `--allow-warnings`)?

justindhill:
You could suppress it at compile-time with a build setting, but unfortunately, that would have the side effect of suppressing _all_ deprecation warnings. I wouldn't recommend it, but that's as specific as you can get with warning suppression.

justindhill:
You can suppress warning emissions within a particular block of code using the preprocessor.

```
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"

// code that touches deprecated APIs here

#pragma clang diagnostic pop
```

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -229,7 +235,11 @@ - (NSDictionary *)liveContext
     });
 
     self.location = !self.location ? [self.configuration shouldUseLocationServices] ? [SEGLocation new] : nil : self.location;
+
+#if TARGET_OS_IOS || (TARGET_OS_MAC && !TARGET_OS_IPHONE)
     [self.location startUpdatingLocation];
+#endif
```
f2prateek: I think  we might want to move this lower down - to right before `return [context copy];`?

f2prateek:
We'll want to remove this entire block:

```
// Check for previous queue/track data in NSUserDefaults and remove if present
         [self dispatchBackground:^{
             if ([[NSUserDefaults standardUserDefaults] objectForKey:SEGQueueKey]) {
                 [[NSUserDefaults standardUserDefaults] removeObjectForKey:SEGQueueKey];
             }
             if ([[NSUserDefaults standardUserDefaults] objectForKey:SEGTraitsKey]) {
                 [[NSUserDefaults standardUserDefaults] removeObjectForKey:SEGTraitsKey];
             }
         }];
```

Analytics/Classes/Internal/SEGSegmentIntegration.m
justindhill: I don't know that I agree. CoreLocation is still available on tvOS, as well as CLLocationManager, but it doesn't provide periodic updates like it does on iOS. Presumably, this is because Apple doesn't anticipate the device moving while it's in use. 😜

justindhill:
Probably only want to remove it on tvOS, right? I guess the purpose of it was to flush data from NSUserDefaults after you rolled back to file-based storage.

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -586,16 +611,26 @@ - (void)applicationWillTerminate
 - (NSMutableArray *)queue
 {
     if (!_queue) {
+#if TARGET_OS_TV
+        _queue = [[NSUserDefaults standardUserDefaults] objectForKey:SEGQueueKey];
```
f2prateek: Hey @justindhill , this will need to be `_queue = [[NSUserDefaults standardUserDefaults] objectForKey:SEGQueueKey]; ?: [[NSMutableArray alloc] init];`. Without that, the queue is always `nil` since nothing has been stored for that key, and none of the payloads are added to the queue, as a result of which nothing is persisted or uploaded.

f2prateek:
@justindhill Yup that was indeed the purpose. Let's remove those LOC just from tvOS for now.

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -586,16 +611,26 @@ - (void)applicationWillTerminate
...
     }
+
     return _queue;
 }
 
 - (NSMutableDictionary *)traits
 {
     if (!_traits) {
+#if TARGET_OS_TV
+        _traits = [[NSUserDefaults standardUserDefaults] objectForKey:SEGTraitsKey];
```
f2prateek: Same as for the queue, this should fallback to ` [[NSMutableDictionary alloc] init]`

`_traits = [[NSUserDefaults standardUserDefaults] objectForKey:SEGTraitsKey]  ?: [[NSMutableDictionary alloc] init];`

Analytics/Classes/Internal/SEGSegmentIntegration.m
justindhill: Ah, good catch

f2prateek:
I tested this with the `?: [[NSMutableDictionary alloc] init]` fallbacks and removing the code that was clearing `SEGQueueKey` and `SEGTraitsKey` and with those changes everything looked to be working fine!

justindhill:
Great, I made those changes in f9eddea!

Analytics/Classes/Internal/SEGSegmentIntegration.m
f2prateek: Ah I see, that makes sense!

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -496,6 +525,8 @@ - (void)reset
     [self dispatchBackgroundAndWait:^{
         [[NSUserDefaults standardUserDefaults] setValue:nil forKey:SEGUserIdKey];
         [[NSUserDefaults standardUserDefaults] setValue:nil forKey:SEGAnonymousIdKey];
+        [[NSUserDefaults standardUserDefaults] setValue:@[] forKey:SEGQueueKey];
```
f2prateek: Should this be only on tvOS as well?

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -496,6 +525,8 @@ - (void)reset
...
+        [[NSUserDefaults standardUserDefaults] setValue:nil forKey:SEGTraitsKey];
```
f2prateek: Why not `@{}` instead of `nil`?

justindhill: No particular reason - to be honest, I copy-pasted these lines out of someone else's fork. I can change it.

Analytics/Classes/Internal/SEGSegmentIntegration.m
justindhill: I think so, yes.

f2prateek:
Thanks, tested manually and merged in #572

PR #578, opened 2016-07-20
Refactor HTTP code and add proxy support.

f2prateek:
This refactors the HTTP interaction in the library + Segment integration. Internally, it moves away from `NSURLConnection` to `NSURLSession`. `SEGAnalyticsRequest` is replaced by `NSURLSessionTask`, and a thin wrapper class `SEGHTTPClient` is introduced to simplify the interaction between the library and the Segment API.

This also adds tests for the networking code by stubbing  the Segment API using https://github.com/luisobo/Nocilla so the behaviour of `SEGHTTPClient` can be verified under different responses from the server.

Publicly, it also adds a `SEGRequestFactory` abstraction that given a URL returns a `NSMutableURLRequest `. The default implementation simply calls `[NSMutableURLRequest requestWithURL:url]`.  It is configurable, and this lets users customize the `NSMutableURLRequest` made by the library. e.g. users could change it to point the SDK to a custom proxy instance.


f2prateek:
CI is failing intermittently for unknown reasons, tested locally and on will's computer.

We've seen green on CI as well https://travis-ci.org/segmentio/analytics-ios/builds/146060030 https://travis-ci.org/segmentio/analytics-ios/builds/146034626 so merging this.

PR #592, opened 2016-08-25
Support client side encryption via SEGCrypto protocol

tonyxiao:
https://segment.atlassian.net/browse/INT-587
Architecture here: https://paper.dropbox.com/doc/Pluggable-Client-Side-Encryption-Architecture-Review-z7spaYemPF4IU5vVlppQL

In order to make this happen we needed to route all disk access through
SEGStorage protocol. There are two concrete implementations, one for file based
and another for user defaults based, depending on OS.

cc @f2prateek @jgershen


tonyxiao:
Tests are here. But unfortunately in Swift, and it's out of scope to setup swift tests in master.
https://github.com/segmentio/analytics-ios/blob/redux/AnalyticsTest/FileStorageTest.swift
https://github.com/segmentio/analytics-ios/blob/redux/AnalyticsTest/UserDefaultsStorageTest.swift
https://github.com/segmentio/analytics-ios/blob/redux/AnalyticsTest/CryptoTest.swift

They do pass of course :)
```
Test Suite 'UserDefaultsStorageTest' started at 2016-08-24 23:01:44.220
Test Case '-[AnalyticsTest.UserDefaultsStorageTest persists_and_loads_data]' started.
Test Case '-[AnalyticsTest.UserDefaultsStorageTest persists_and_loads_data]' passed (0.001 seconds).
Test Case '-[AnalyticsTest.UserDefaultsStorageTest persists_and_loads_string]' started.
Test Case '-[AnalyticsTest.UserDefaultsStorageTest persists_and_loads_string]' passed (0.001 seconds).
Test Case '-[AnalyticsTest.UserDefaultsStorageTest persists_and_loads_array]' started.
Test Case '-[AnalyticsTest.UserDefaultsStorageTest persists_and_loads_array]' passed (0.001 seconds).
Test Case '-[AnalyticsTest.UserDefaultsStorageTest persists_and_loads_dictionary]' started.
Test Case '-[AnalyticsTest.UserDefaultsStorageTest persists_and_loads_dictionary]' passed (0.001 seconds).
Test Case '-[AnalyticsTest.UserDefaultsStorageTest should_work_with_crypto]' started.
Test Case '-[AnalyticsTest.UserDefaultsStorageTest should_work_with_crypto]' passed (0.034 seconds).
Test Case '-[AnalyticsTest.UserDefaultsStorageTest should_work_with_namespace]' started.
Test Case '-[AnalyticsTest.UserDefaultsStorageTest should_work_with_namespace]' passed (0.034 seconds).
Test Suite 'UserDefaultsStorageTest' passed at 2016-08-24 23:01:44.293.
	Executed 6 tests, with 0 failures (0 unexpected) in 0.071 (0.073) seconds
Test Suite 'FileStorageTest' started at 2016-08-24 23:01:42.693
Test Case '-[AnalyticsTest.FileStorageTest creates_folder_if_none_exists]' started.
Test Case '-[AnalyticsTest.FileStorageTest creates_folder_if_none_exists]' passed (0.001 seconds).
Test Case '-[AnalyticsTest.FileStorageTest persists_and_loads_data]' started.
Test Case '-[AnalyticsTest.FileStorageTest persists_and_loads_data]' passed (0.002 seconds).
Test Case '-[AnalyticsTest.FileStorageTest persists_and_loads_string]' started.
Test Case '-[AnalyticsTest.FileStorageTest persists_and_loads_string]' passed (0.001 seconds).
Test Case '-[AnalyticsTest.FileStorageTest persists_and_loads_array]' started.
Test Case '-[AnalyticsTest.FileStorageTest persists_and_loads_array]' passed (0.002 seconds).
Test Case '-[AnalyticsTest.FileStorageTest persists_and_loads_dictionary]' started.
Test Case '-[AnalyticsTest.FileStorageTest persists_and_loads_dictionary]' passed (0.002 seconds).
Test Case '-[AnalyticsTest.FileStorageTest saves_file_to_disk_and_removes_from_disk]' started.
Test Case '-[AnalyticsTest.FileStorageTest saves_file_to_disk_and_removes_from_disk]' passed (0.001 seconds).
Test Case '-[AnalyticsTest.FileStorageTest should_be_binary_compatible_with_old_SDKs]' started.
Test Case '-[AnalyticsTest.FileStorageTest should_be_binary_compatible_with_old_SDKs]' passed (0.002 seconds).
Test Case '-[AnalyticsTest.FileStorageTest should_work_with_crypto]' started.
Test Case '-[AnalyticsTest.FileStorageTest should_work_with_crypto]' passed (0.034 seconds).
Test Suite 'FileStorageTest' passed at 2016-08-24 23:01:42.752.
	 Executed 8 tests, with 0 failures (0 unexpected) in 0.045 (0.059) seconds
Test Suite 'CryptoTest' started at 2016-08-24 23:01:42.575
Test Case '-[AnalyticsTest.CryptoTest encrypts_and_decrypts_data]' started.
Test Case '-[AnalyticsTest.CryptoTest encrypts_and_decrypts_data]' passed (0.034 seconds).
Test Case '-[AnalyticsTest.CryptoTest fails_for_incorrect_password]' started.
Test Case '-[AnalyticsTest.CryptoTest fails_for_incorrect_password]' passed (0.034 seconds).
Test Case '-[AnalyticsTest.CryptoTest fails_for_incorrect_iv_and_sault]' started.
Test Case '-[AnalyticsTest.CryptoTest fails_for_incorrect_iv_and_sault]' passed (0.033 seconds).
Test Suite 'CryptoTest' passed at 2016-08-24 23:01:42.678.
	 Executed 3 tests, with 0 failures (0 unexpected) in 0.101 (0.103) seconds
```

Analytics/Classes/Internal/SEGCrypto.h
```diff
@@ -0,0 +1,15 @@
+//
+//  SEGCrypto.h
+//  Analytics
+//
+//  Copyright © 2016 Segment. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@protocol SEGCrypto <NSObject>
```
f2prateek: This would be public so users can supply their own implementation, yeah?

Analytics/Classes/Internal/SEGFileStorage.m
```diff
@@ -0,0 +1,164 @@
...
+    if (self.crypto) {
+        NSData *encryptedData = [self.crypto encrypt:data];
+        [encryptedData writeToURL:url atomically:YES];
+    } else {
+        [data writeToURL:url atomically:YES];
+    }
+    
+    NSError *error = nil;
+    if (![url setResourceValue:@YES
+                        forKey:NSURLIsExcludedFromBackupKey
```
f2prateek: Not sure if this is the best place for this - what if we do want some of the keys to be backed up and shared across devices? I've considered doing so for anonymousId and userId pending some testing on multiple devices.

Analytics/Classes/Internal/SEGCrypto.h
tonyxiao: Yep.

Analytics/Classes/Internal/SEGFileStorage.m
tonyxiao: If we do that responsibility should still live inside the storage implementation, but we can just provide an API on top like
`setBackup:(BOOL)backup forKey:(NSString *)key`

tonyxiao: Pretty simple to add if / when we do need it.

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -630,32 +619,17 @@ - (NSUInteger)maxBatchSize
...
 {
-    return [[NSUserDefaults standardUserDefaults] valueForKey:SEGUserIdKey] ?: [[NSString alloc] initWithContentsOfURL:self.userIDURL encoding:NSUTF8StringEncoding error:NULL];
+    return [[NSUserDefaults standardUserDefaults] valueForKey:SEGUserIdKey] ?: [self.analytics.storage stringForKey:kSEGUserIdFilename];
 }
 
 - (void)persistQueue
 {
 #if TARGET_OS_TV
-    [[NSUserDefaults standardUserDefaults] setValue:[self.queue copy] forKey:SEGQueueKey];
+    [self.analytics.storage setArray:[self.queue copy] forKey:SEGQueueKey];
```
tonyxiao: @f2prateek this is pretty in-elegant because the name of keys are different depending on whether we are using user defaults or files. I'd be interested in getting a proper migration system setup where we can make this consistent so no need to do these crazy `#if` statements.

f2prateek: I think these macros should be handled in the SEGFileStorage class - I thought that was one the reasons of the SEGFileStorage abstraction. 

It could probably be just called SEGStorage now that I think about it since it'd deal with NSUserDefaults and files.

f2prateek:
We should definitely port the tests here either in Swift or just with what's already setup https://github.com/segmentio/analytics-ios/blob/master/Example/Tests/SEGHTTPClientTests.m.

f2prateek:
also the PR should be against the `dev` branch https://github.com/segmentio/analytics-ios/blob/master/CONTRIBUTING.md

Analytics/Classes/Internal/SEGSegmentIntegration.m
tonyxiao: It's not file based storage though. so doesn't make sense to handle it in file storage class.

tonyxiao:
@f2prateek changed based to `dev`. We should have a separate PR to support testing in Swift rather than expanding this scope of this PR.

Analytics/Classes/Internal/SEGSegmentIntegration.m
tonyxiao: There is a single macro already at `SEGAnalytics` initialization time that determines whether to use `SEGFileStorage` or `SEGUserDefaultsStorage` depending on the OS. Our issue is that we use in-consistent keys.

tonyxiao: Not sure if it's worth it to have yet another abstraction wrapping these two forms of storages at the moment.

f2prateek:
@tonyxiao makes sense about testing in Swift, I don't think this should be released without adding some tests so we can just write them in objective c for now.

Analytics/Classes/Internal/SEGSegmentIntegration.m
f2prateek: Oh sorry I think I misunderstood the file storage stuff. It looks like there is already a SEGStorage that switches implementations for tvOS.

Analytics/Classes/Internal/SEGUserDefaultsStorage.h
```diff
@@ -0,0 +1,20 @@
...
+//  Created by Tony Xiao on 8/24/16.
+//  Copyright © 2016 Segment. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "SEGStorage.h"
+
+@interface SEGUserDefaultsStorage : NSObject <SEGStorage>
+
+@property (nonatomic, strong, nullable) id<SEGCrypto> crypto;
```
f2prateek: It seems a bit leaky that the the storage mechanism knows about encryption. Also causes fair bit of duplicated code b/w the two storage implementations to deal with crypto.

I think the encryption should happen before call the storage methods, and decryption after.

The usage could be like this:

```
NSData *encrypted = [crypto encrypt:data];
[storage setData:encrypted forKey:key];
```

```
NSData *encrypted = [storage dataForKey:key];
NSData *decrypted = [crypto decrypt:encrypted];
```

tonyxiao: This makes it difficult to be backward compatible with old storage mechanisms. In particular for user default based storage we don't always storage `NSData` objects, we directly store `NSString` and etc.

tonyxiao:
Is it worth the effort to port the tests considering we'll actually have real Swift setup soon? Most of our current code also does not contain test coverage.

Analytics/Classes/Internal/SEGUserDefaultsStorage.h
f2prateek: Hmmm...  I see. I think that's ok, the user defaults code is only for tvOS and is only been available for a single release, and it gives us a window to change some of the implementation and we should use it now rather than be stuck with a legacy implementation.

f2prateek:
I think it definitely is worth the effort of having tests for new features. It doesn't have to be a full end to end test, but at least have unit tests like https://github.com/segmentio/analytics-ios/blob/master/Example/Tests/SEGHTTPClientTests.m. We haven't had time to bring back the removed tests, but should at least keep adding tests for new components.

Analytics/Classes/Internal/SEGCrypto.h
f2prateek: Cool so let's move it out of the `Internal` folder?

tonyxiao:
Hmm between re-writing tests in Objc and setting up tests in swift I'd rather just set up tests in Swift so no effort is wasted. Maybe I should get that in first or part of this PR.

Analytics/Classes/Internal/SEGUserDefaultsStorage.h
f2prateek: Although I think this shouldn't be a blocker for the change.

f2prateek:
I don't think it should be that much work to port the tests, I'm happy to do it myself once this branch is ready (it might already be)

Analytics/Classes/Internal/SEGUserDefaultsStorage.h
tonyxiao: @f2prateek are we ok not worrying about migration on tvOS since last version?

f2prateek: @tonyxiao Possibly, given support has been very recently added (just since the last release) I'm not too worried about migrating it cleanly for tvOS, if it helps us implementing this.

But as I said, doesn't need to be a blocker, should be ok as is and we can refine it later since the FileStorage class is internal.

tonyxiao:
It's all good. I'll take a stab at getting the Swift tests running in this branch, shouldn't to too bad. (After all, we have it already elsewhere)

tonyxiao:
@f2prateek all comments addressed.

Analytics/Classes/Crypto/SEGAES256Crypto.h
```diff
@@ -0,0 +1,23 @@
...
+
+@property (nonatomic, readonly, nonnull) NSString *password;
+@property (nonatomic, readonly, nonnull) NSData *salt;
+@property (nonatomic, readonly, nonnull) NSData *iv;
+
+- (instancetype _Nonnull)initWithPassword:(NSString * _Nonnull)password salt:(NSData * _Nonnull)salt iv:(NSData * _Nonnull)iv;
+// Convenient shorthand. Will randomly generate salt and iv.
+- (instancetype _Nonnull)initWithPassword:(NSString * _Nonnull)password;
+
++ (NSData * _Nonnull)randomDataOfLength:(size_t)length;
```
f2prateek: Does this need to be public?

Analytics/Classes/Crypto/SEGAES256Crypto.h
```diff
@@ -0,0 +1,23 @@
...
+//
+//  Copyright © 2016 Segment. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "SEGCrypto.h"
+
+@interface SEGAES256Crypto : NSObject <SEGCrypto>
+
+@property (nonatomic, readonly, nonnull) NSString *password;
```
f2prateek: Do these properties  need to be public?

Example/Tests/UserDefaultsStorageTest.swift
```diff
@@ -0,0 +1,103 @@
...
+
+import Quick
+import Nimble
+import Analytics
+
+class UserDefaultsStorageTest : QuickSpec {
+  override func spec() {
+    var storage : SEGUserDefaultsStorage!
+    beforeEach {
+//      let crypto = SEGAES256Crypto(password: "thetrees")
```
f2prateek: Could be removed if unused?

Analytics/Classes/Crypto/SEGAES256Crypto.h
tonyxiao: Yes, user will need to persist password, iv, and salt in order to properly initialize the crypto object next time.

Analytics/Classes/Crypto/SEGAES256Crypto.h
tonyxiao: Yes, it's a convenience method for people to generate `salt` and `iv` and possibly `password`.

Analytics/Classes/Crypto/SEGAES256Crypto.h
tonyxiao: This is especially true in the case where iv and salt are auto-generated.

Example/Tests/UserDefaultsStorageTest.swift
tonyxiao: It was actually kind of useful because it allowed changing the storage configuration and manually re-run the tests. Now we do have tests for multiple storage configuration, but ideally I'd love to have parametric test, where the same test suite is run but with different parameters. Not sure how to do them, which is the only reason I added a few sanity tests at the bottom (should work with namespace, should work with crypto, etc).

f2prateek:
LGTM

PR #597, opened 2016-09-13
Add option to enable/disable bluetooth collection.

f2prateek:
Also disables bluetooth collection by default.

### DIFF
--- a/Analytics/Classes/Internal/SEGSegmentIntegration.m
+++ b/Analytics/Classes/Internal/SEGSegmentIntegration.m
@@ -99,7 +99,9 @@ static BOOL GetAdTrackingEnabled()
         self.httpClient = analytics.httpClient;
         self.apiURL = [NSURL URLWithString:@"https://api.segment.io/v1/import"];
         self.userId = [self getUserId];
-        self.bluetooth = [[SEGBluetooth alloc] init];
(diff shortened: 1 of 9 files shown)


f2prateek:
cc @tonyxiao

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -236,7 +238,7 @@ - (NSDictionary *)liveContext
     context[@"network"] = ({
         NSMutableDictionary *network = [[NSMutableDictionary alloc] init];
 
-        if (self.bluetooth.hasKnownState)
+        if (self.bluetooth && self.bluetooth.hasKnownState)
```
tonyxiao: FWIW in Objc if `self.bluetooth` is `nil` then `self.bluetooth.hasKnownState` will evaluate to `NO` automatically. It's kind of an idiomatic objc thing, but of course both way works.

tonyxiao:
lgtm

Analytics/Classes/Internal/SEGSegmentIntegration.m
f2prateek: Yeah I was considering not adding the `self.bluetooth` check - I don't have a strong preference either way so I ended up adding it.

PR #600, opened 2016-09-17
Converting tests to swift

tonyxiao:
@f2prateek I think we should standardize on using Swift for writing tests so we don't need tests in both ObjC and Swift.


Example/Tests/HTTPClientTest.swift
```diff
@@ -0,0 +1,114 @@
...
+        stubRequest("GET", "https://cdn.segment.com/v1/projects/foo/settings")
+          .withHeaders(["Accept-Encoding" : "gzip" ])
+          .andReturn(200)
+          .withHeaders(["Content-Type" : "application/json"])
+          .withBody("{\"integrations\":{\"Segment.io\":{\"apiKey\":\"foo\"}},\"plan\":{\"track\":{}}}")
+
+        var done = false
+        let task = client.settingsForWriteKey("foo", completionHandler: { success, settings in
+          expect(success) == true
+          expect(settings) == [
```
f2prateek: oh nice - this works coz of operator overloading?

f2prateek:
LGTM

PR #604, opened 2016-09-20
First implementation of Middlewares

tonyxiao:
cc @f2prateek

Mostly just moved a ton of core integration management logic into SEGIntegraionsManager
and let SEGAnalytics delegate to it via the middleware pipeline.
* SEGAnalytics is still responsible for "sources". Basically automatically invoking
track and other calls based on application state and other environment factors.

We now have the ability to make split things up into much more logical and composable
pieces in the middleware pipeline.

If one were to not care about bundled integrations at all, we can easily make SEGSegmentIntegration
simply conform to the SEGMiddleware protocol. And in place of SEGIntegrationsManager simply
subtitute the direct SEGSegmentIntegration and now you have yourself a much more simplified
pipeline without all the black magic which is integrations manager. 
* (If we want even filtering even for segment only btw, we can easily extract out the event 
  filtering piece into its own middleware :) composability ftw)

We probably should make SEGSegmentIntegration (and all SEGIntegrations for that matter)
conform to the SEGMiddleware protocol (via some kind of adapter class initially, and natively
later). 

SEGIntegrationsManager can then simply be a fanout middleware that dispatches events to a whole
bunch of middlewares in parallel.

Testing this in a live app now and see if it works


tonyxiao:
Ok tested in live app and is working!

tonyxiao:
Btw while this set of commits is a bit scary I mostly just moved files around and also pod changes took up a whole bunch of changes :(

f2prateek:
LGTM!

Issue #476, opened 2015-11-19
Ability to change write key on the fly?


DimaVartanian:
Looks like the Swift library has the ability to do this actually. Would it be possible to return a SEGAnalytics instance and allow the app to save and use that rather than a singleton? Our beta/enterprise releases give users the ability to switch between our production and sandbox environments and it would be nice to change the write key based on that to allow events to go to the right place (as we are also using these events to identify users elsewhere).

f2prateek:
Not currently. I would like to do this eventually, especially since we also do it in the Android library. It's not terribly hard, but the problem comes from bundled integrations, and since very few of them have the ability to have multiple instances, having multiple instances of the the segment analytics client but a single instance of say the Mixpanel SDK is likely to cause problems. The Swift library doesn't have any bundled integrations, so it can get away with it.

f2prateek:
Tracking internally in JIRA as well — https://segment.atlassian.net/browse/LIB-135

pavermakov:
Any updates on it?

alanjcharles:
Hi @pavermakov this isn't something we will likely pursue for this library, I'm afraid. I would definitely recommend checking out our new analytics-swift pilot for a more programmatic approach to your analytics implementation: 

https://github.com/segmentio/analytics-swift

PR #609, opened 2016-09-29
Fix for retain cycle of SEGSegmentIntegration and its flush timer

AndyStyles:
Releasing instances of SEGAnalytics causes a crash as SEGSegmentIntegration contains a retain cycle with its NSTimer used to trigger flushes. Due to the fact NSTimer keeps a strong reference to its target object and the timer itself is retained by the run loop, a weak referenced target object is required.

The SEGTimerProxy holds the actual target object weakly and forwards the timer selector to the real object. This breaks the retain cycle and allows SEGSegmentIntegration to be deallocated, upon deallocation of SEGSegmentIntegration the NSTimer can be invalidated completing the clean up.

### DIFF
--- a/Analytics/Classes/Internal/SEGSegmentIntegration.m
+++ b/Analytics/Classes/Internal/SEGSegmentIntegration.m
@@ -9,6 +9,7 @@
 #import "SEGLocation.h"
 #import "SEGHTTPClient.h"
 #import "SEGStorage.h"
+#import "SEGTimerProxy.h"
 
 #if TARGET_OS_IOS
 #import <CoreTelephony/CTCarrier.h>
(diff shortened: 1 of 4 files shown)


f2prateek:
It seems reasonable to me - but I need to review the implementation a bit more.

I do have a question though - why are you releasing instances of SEGAnalytics? It's meant to be initialized during application launch as a singleton and live for the lifetime of your app. While we definitely want to improve this design in the future - I'm not entirely sure what use cases this enables.

AndyStyles:
Thanks for the response - basically we need to be able to change the write key while the app is running. Because of this we were looking to avoid using the singleton you provide and user our own wrapper that holds a SEGAnalytics instance, when we need to change the write key we init a new SEGAnalytics and release the old one. Or have I missed something and this is possible some other way?

f2prateek:
@AndyStyles That's definitely a reasonable request and something we're planning on supporting with the next version. Unfortunately currently you can't really change the writeKey on the fly. It *might* work, but will have some odd side effects since our on disk caches are not namespaced. e.g. an analytics instance with the key `foo` will use the same queue file as the analytics instance with the key `bar`.

AndyStyles:
@f2prateek Ah thanks for the information, for now we can avoid changing the key on the fly while we wait for the functionality to be added. Feel free to close this PR if you do not see a scenario where a SEGAnalytics instance would be released. Cheers

f2prateek:
I'll close this PR, and we'll track progress in https://github.com/segmentio/analytics-ios/issues/476

PR #610, opened 2016-10-05
Remove bluetooth collection entirely.

tonyxiao:
cc @f2prateek. Until we have time to get a better handle on bluetooth usage issue with iOS 10 it's best to be conservative and completely remove bluetooth to not block customers.


tonyxiao:
Cherry-picking onto master in a separate PR here https://github.com/segmentio/analytics-ios/pull/611

f2prateek:
LGTM

PR #611, opened 2016-10-05
[Hotfix] Remove bluetooth collection entirely.

tonyxiao:
cc @f2prateek. Until we have time to get a better handle on bluetooth usage issue with iOS 10 it's best to be conservative and completely remove bluetooth to not block customers.

This is a hotfix to address the immediate issue without waiting for regular release train


f2prateek:
LGTM

PR #612, opened 2016-10-05
Remove location collection entirely

tonyxiao:
similar story to here https://github.com/segmentio/analytics-ios/pull/610. We'll figure out a good overall strategy for adding this back.

PR #613, opened 2016-10-05
[hotfix] Remove location collection entirely

tonyxiao:
non-hotfix version here https://github.com/segmentio/analytics-ios/pull/612
Similar story to https://github.com/segmentio/analytics-ios/pull/611

We'll add it back as part of vNext when ready

cc @f2prateek


f2prateek:
LGTM1

PR #615, opened 2016-10-10
Initialize storage before generating anonymous ID.

f2prateek:
Since 3.5.0 (https://github.com/segmentio/analytics-ios/pull/592/files), a regression was introduced as part of the crypto changes
that would cause a new anonymous ID to be generated for every launch of
the app.

This change fixes that regression.


tonyxiao:
@f2prateek if we used Swift for the core library this issue would have been caught by the compiler. Unfortunately we are not able to use Swift in the core library, but are there other ways this type of issues (nil pointer access silently dropped) could have been caught by maybe using nullability indicator? 

Unfortunately sending a message to nil is one of the most common kind of bugs on iOS :( http://stackoverflow.com/questions/156395/sending-a-message-to-nil

Issue #616, opened 2016-10-11
Crash accessing "referrer" on liveContext after calling Segment's continueUserActivity()


andrewloyola:
We've got some code that handles universal links (via branch), and we're seeing a consistent crash with Analytics 3.3.0 and later releases. This is on iOS 10, both beta and public versions.

We receive the AppDelegate `continueUserActivity` call when handling a universal link. Inside the delegate call after calling `SEGAnalytics.sharedAnalytics().continueUserActivity()` which saves the referring URL, the next track call crashes in SEGSegmentIntegration.m when trying to set `context[@"referrer"] ` when accessing `liveContext`.  Commenting out the `SEGAnalytics.sharedAnalytics().continueUserActivity()` call makes the crashes stop.

So far seems to be an issue with setting the referrer dictionary with activity.webpageURL from the NSUserActivity object.  When this is accessed later, it is nonnull but it is also not a Dictionary when looking at the debugger. 

Curious to see if anyone else is seeing this behavior?  It doesnt seem to be related to the URL itself; at least i've tried substituted a new NSUserActivity with a known good url to test that part out.

f2prateek:
Can you share the stacktrace?

f2prateek:
Also a reproducible test case would be great!

f2prateek:
@andrewloyola from a quick guess, it looks like [`activity.WebpageURL` might be nil](https://github.com/segmentio/analytics-ios/blob/5cad6f5dadb0f1304a12248e15ced20335204f8b/Analytics/Classes/Internal/SEGSegmentIntegration.m#L385).

I'm not entirely sure why it would be `nil`. We [check if the activityType is `NSUserActivityTypeBrowsingWeb`](https://github.com/segmentio/analytics-ios/blob/5cad6f5dadb0f1304a12248e15ced20335204f8b/Analytics/Classes/Internal/SEGSegmentIntegration.m#L383), which according to [docs](https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/Handoff/AdoptingHandoff/AdoptingHandoff.html) should contain the  `webpageURL` property.

andrewloyola:
sure, attaching a screenshot at least of the stacktrace while I figure out how to print the equivalent.
![crash](https://cloud.githubusercontent.com/assets/1688758/19288712/567d46de-8fd5-11e6-99f6-a0cd887e87f8.png)

andrewloyola:
[...]

PR #617, opened 2016-10-17
Fixed crash due to the referrer dictionary being `assign`

AriX:
`assign` is a synonym for `unsafe_unretained` in ARC, so if the dictionary is released the pointer points to a deallocated object and the app will crash when it is used!


f2prateek:
Thanks, this was raised in #616, and we have a fix in https://github.com/segmentio/analytics-ios/commit/82d8495bc359897e2a7380752df4856e5cbf5887 as well. Planning to have that merged in the next release, sorry for the trouble!

AriX:
Ahh, got it - sounds great!

f2prateek:
Closed by #621

PR #618, opened 2016-10-18
Upgrade tests from Swift 2 -> 3 and work with Xcode 8

tonyxiao:
cc @f2prateek 

In addition to being a good practice (to upgrade code to latest), this is also needed to make tests work with xcode 8.

Try looking at the individual commits for a cleaner view without crazy Pods changes.

https://github.com/segmentio/analytics-ios/pull/618/commits


f2prateek:
LGTM

PR #622, opened 2016-10-22
Re-dispatch callbacks from SEGHTTPClient onto correct queues before executing logic

tonyxiao:
Likely should help fix issue. 

https://segment.atlassian.net/browse/EPD-142
https://segment.atlassian.net/browse/EPD-198

In a separate (not hot fix PR) we should make it easier to have correct threading logic
by letting `SEGHTTPClient` take in a dispatch queue upon initialization

cc @f2prateek @ladanazita


tonyxiao:
Please ignore circle build failure. We don't have circle.yml for master yet. Just watch if travis succeeds.

f2prateek:
LGTM

PR #628, opened 2016-10-29
End background tasks as soon as flushing is complete.

maxmeyers:
This fixes what I consider to be a critical bug, reproducible in the Example app.

`SEGSegmentIntegration` currently does not call `UIApplication#endBackgroundTask` until the expiration handler from `beginBackgroundTask` is called, which means the app will remaining running open in the background until the expiration handler is called. `SEGSegmentIntegration` should behave as a good citizen and end the background task as soon as it no longer needs.

To reproduce:
1. Run the Example app in the debugger
2. Press Command-Shift-H to background the app without killing its process.
3. Pause the debugger and type in `po [(UIApplication*)[UIApplication sharedApplication] backgroundTimeRemaining]`
4. Note the result is a large number over 100, indicating the app will remain active for several minutes.

This change fixes it by calling `endBackgroundTask` from the three locations in the path taken after backgrounding the app. After making the change, the result of the print out from step 3 above is `1.7976931348623157E+308`, indicating the app has been suspended.

### DIFF
--- a/Analytics/Classes/Internal/SEGSegmentIntegration.m
+++ b/Analytics/Classes/Internal/SEGSegmentIntegration.m
@@ -471,7 +471,6 @@ static CTTelephonyNetworkInfo *_telephonyNetworkInfo;
         }
         if (self.batchRequest != nil) {
             SEGLog(@"%@ API request already in progress, not flushing again.", self);
-            [self endBackgroundTask];
             return;
         }
 
--- a/Analytics/Classes/Internal/SEGSegmentIntegration.m
+++ b/Analytics/Classes/Internal/SEGSegmentIntegration.m
@@ -539,6 +539,7 @@ static CTTelephonyNetworkInfo *_telephonyNetworkInfo;
         if (retry) {
             [self notifyForName:SEGSegmentRequestDidFailNotification userInfo:batch];
             self.batchRequest = nil;
+            [self endBackgroundTask];
             return;
         }
 
--- a/Analytics/Classes/Internal/SEGSegmentIntegration.m
+++ b/Analytics/Classes/Internal/SEGSegmentIntegration.m
@@ -466,10 +466,12 @@ static CTTelephonyNetworkInfo *_telephonyNetworkInfo;
     [self dispatchBackground:^{
         if ([self.queue count] == 0) {
             SEGLog(@"%@ No queued API calls to flush.", self);
+            [self endBackgroundTask];
             return;
         }
         if (self.batchRequest != nil) {
             SEGLog(@"%@ API request already in progress, not flushing again.", self);
+            [self endBackgroundTask];
             return;
         }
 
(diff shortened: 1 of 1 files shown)


Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -466,10 +466,12 @@ - (void)flushWithMaxSize:(NSUInteger)maxBatchSize
     [self dispatchBackground:^{
         if ([self.queue count] == 0) {
             SEGLog(@"%@ No queued API calls to flush.", self);
+            [self endBackgroundTask];
             return;
         }
         if (self.batchRequest != nil) {
             SEGLog(@"%@ API request already in progress, not flushing again.", self);
+            [self endBackgroundTask];
```
f2prateek: There is already a flush in progress (which will take care of ending the background task), so I don't think we should end it here.

f2prateek:
cc @tonyxiao who will approve and merge these changes since I'm out next week.

Analytics/Classes/Internal/SEGSegmentIntegration.m
maxmeyers: Ah, that makes sense.

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -466,6 +466,7 @@ - (void)flushWithMaxSize:(NSUInteger)maxBatchSize
     [self dispatchBackground:^{
         if ([self.queue count] == 0) {
             SEGLog(@"%@ No queued API calls to flush.", self);
+            [self endBackgroundTask];
```
tonyxiao: @maxmeyers what's the rationale for this line? Why aren't the other two `endBackgroundTask` calls sufficient?

maxmeyers: The `return;` on the following line prevents the batch request from ever starting. The other two `endBackgroundTask` calls are in the completion handler of the batch request.

tonyxiao:
Thanks for the contribution @maxmeyers. Merging

maxmeyers:
What's the ETA on getting this fix released? CONTRIBUTING.md says there's a release every second Wednesday.

tonyxiao:
@maxmeyers We have a pretty major release coming so doing a bit more testing than usual. However I will cherry-pick this change and make a hotfix release on top of master today.

maxmeyers:
@tonyxiao It looks like this never made it into a release. Any plans for that?

f2prateek:
@maxmeyers This should have been part of the 3.6.0-beta release - are you seeing otherwise?

maxmeyers:
Oh, I see. I ask because on Nov. 4th Tony said there would be a hot fix release with this change. We'd rather not ship a beta release into production, of course.

maxmeyers:
@tonyxiao @f2prateek Any plans to ship a hot fix release with this change? Or should I wait for 3.6.0?

tonyxiao:
@maxmeyers Currently there isn't a plan to cut a hotfix release for this particular issue. 

That being said 3.6.0-beta is actually pretty stable. We've done more testing on this particular version than all of our other versions. The only reason it's beta right now is because we are starting a new release process that's extra cautious and have multiple channels (Similar to chrome canary, beta, stable, etc). This is a new concept so that's why the first release out of it is labeled `beta` even if it's pretty stable.

Issue #625, opened 2016-10-25
tvOS support througth carthage 


BProg:
building using `carthage update analytics-ios --platform tvOS`
gives error: `Dependency "analytics-ios" has no shared framework schemes for any of the platforms: tvOS`

If use binary tagged at 3.5.2 which is built for iOS, xcode gives me error : 
`URGENT: building for tvOS simulator, but linking against dylib (/Users/admin/Workspace/tvOSApp/Analytics.framework/Analytics) built for iOS. Note: This will be an error in the future.`

f2prateek:
tvOS is supported via CocoaPods, but support via Carthage isn't on our immediate roadmap. Happy to accept PRs that add support in the meantime though!

PR #632, opened 2016-11-14
Add tvOS for deployment target

BProg:
issue: #625

This adds possibility to build the framework for apple tv using `carthage update --platform tvOS analytics-ios`

### DIFF
--- a/Analytics.xcodeproj/project.pbxproj
+++ b/Analytics.xcodeproj/project.pbxproj
@@ -707,7 +707,11 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = com.segment.Analytics;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				SDKROOT = iphoneos;
 				SKIP_INSTALL = YES;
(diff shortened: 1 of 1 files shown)


BProg:
Hi there. Any comments on this? I could rebase it and drop commit 1 and 2 but I am waiting for feedback first.

f2prateek:
I can't find any docs on this on Carthage - could you point us to where this approach is documented?

In case there are no docs about this,  have you had a chance to test this locally? If so, we can do a manual test ourselves as well to verify.

cc @tonyxiao

BProg:
Carthage only needs the framework to be shared. This is the only condition. 
Here I changed the Analytics framework to be cross-platform, so it can be build for apple tv and ios without need of another target for apple TV.

BProg:
Yes I tested and it works.

BProg:
@f2prateek - I saw circle ci is using xcode 7.0.1 for testing, but that xcode does even have a target for apple tv. Do you use xcode 7.0.1 in development? If yes, why?

f2prateek:
I used 8.1 on my computer, probably just haven't gotten around to updating it in circle.

BProg:
If the command `carthage build --no-skip-current` will be run in the project root folder, the folder  `Carthage/Build/` will contain now 2 folders : `iOS` and `tvOS`.

tonyxiao:
lgtm

PR #633, opened 2016-11-15
Workaround to handle IOS 8.3 bug where Content-type of application/x-…

tyliaw:
A workaround for an IOS 8.3 issue that was causing POSTs to api.segment.io to be made with Content-type of application/x-www-form-urlencoded; instead of application/json.


tyliaw:
Additional info can be seen here: http://openradar.appspot.com/20498383

f2prateek:
Could you add a comment in the code explaining why this is needed so when we take a look at it a few years down the line it's easily visible why?

f2prateek:
LGTM otherwise, cc @tonyxiao

tonyxiao:
Seems like it's an iOS bug. LGTM also pending CI passing.

PR #639, opened 2016-11-28
Removing travis because it's too slow relative to circle

tonyxiao:
There's really no need to have both of them running at the same time. Travis builds routinely gets queued 1+ hour before they actually even start. By which time circle would have long been finished. 

Downside of keeping them both is that the extra status check from travis can make it appear that we merge before CI passing, when in fact that's not the case. It's also more work to keep maintaining `.travis.yml`

cc @f2prateek @ladanazita


f2prateek:
SGTM

PR #640, opened 2016-11-28
Remove checked in Pods to simplify github PR

tonyxiao:
Install pods on CI instead prior to running tests

cc @ladanazita @f2prateek


f2prateek:
LGTM - what's the plan for Carthage here?

tonyxiao:
Will follow up with Carthage build steps in the next PR. Keeping pods change separate in this PR due to the massive # of files updated.

PR #641, opened 2016-11-29
Using a standalone xcodeproj for Analytics.framework development

tonyxiao:
* Separate xcodeproj and Podfile for demonstrating Cococapods example
* Should already work with Carthage right now. Will demonstrate in a separate PR carthage support

cc @f2prateek @ladanazita


f2prateek:
LGTM

PR #644, opened 2016-11-29
Consolidate Cocoapods example from /Example into /Examples/CocoapodsExample

tonyxiao:
Final example consolidated. Now we have examples for all three different officially supported setup flows
cc @f2prateek @ladanazita 

screen shot 2016-11-28 at 8 51 50 pm


tonyxiao:
Codecov seems to have some issues with their checks badge. It's actually fine.
image

f2prateek:
LGTM

tonyxiao:
@f2prateek @ladanazita thinking of automating a makefile to test that all three library build options continue to succeed. Let me know if you think it's worth investing in something like that (so we won't have to do manual checks)

PR #647, opened 2016-11-29
Make API a lot easier to use in swift

tonyxiao:
And help compiler catch bugs via generics / nullability indicators

cc @segmentio/mobile

### DIFF
--- a/Analytics/Classes/SEGAnalyticsConfiguration.h
+++ b/Analytics/Classes/SEGAnalyticsConfiguration.h
@@ -8,7 +8,7 @@
 
 #import <Foundation/Foundation.h>
 
-typedef NSMutableURLRequest * (^SEGRequestFactory)(NSURL *);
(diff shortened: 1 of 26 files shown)


Analytics/Classes/SEGAnalyticsConfiguration.h
```diff
@@ -89,21 +89,21 @@ typedef NSMutableURLRequest * (^SEGRequestFactory)(NSURL *);
 /**
  * Dictionary indicating the options the app was launched with.
  */
-@property (nonatomic, assign) NSDictionary *launchOptions;
+@property (nonatomic, strong, nullable) NSDictionary *launchOptions;
```
tonyxiao: @f2prateek calling particular attention to this one. it used to be `assign` which isn't right.

tonyxiao:
codecov is making this PR look bad (due to the ✘ mark) when in fact it doesn't really affect code coverage. Know any workaround @segmentio/mobile

Issue #648, opened 2016-11-29
Leak observed when setupWithConfiguration


ionutivan:
Steps to replicate:
* add below code to AppDelegate:
`      
let configuration = SEGAnalyticsConfiguration(writeKey: ***)        
        SEGAnalytics.setupWithConfiguration(configuration)`
* profile the app and the leak happens in 
`-[SEGHTTPClient settingsForWriteKey:completionHandler:]`

tonyxiao:
@ionutivan I took a quick look and not sure what you are referring to. What do you mean by leak? Which object is retained when you expect it to be released?

ionutivan:
I used Instruments to profile my app and I noticed there is a leak here. So I don't know exactly why it is happening, I just saw it is happening. @tonyxiao

tonyxiao:
Typically a leak means that memory usage keeps going up but does not come down. Over time the app gets killed by the OS for too much memory usage. However this is a one-off request so that's why I'm not sure what you mean by leak here. Could you attach a screenshot of the instrument view that shows the leak you're referring to? @ionutivan

ionutivan:
<img width="1396" alt="screen shot 2016-11-29 at 6 07 04 pm" src="https://cloud.githubusercontent.com/assets/1093056/20722592/18a09dec-b65f-11e6-94e0-56adc2bd7120.png">
@tonyxiao

ionutivan:
Is there anything more I can do to help with this issue @tonyxiao ?

marsepu:
I posted a new PR because I found the same problem. https://github.com/segmentio/analytics-ios/pull/652

marsepu:
@tonyxiao && @ionutivan ^^

tonyxiao:
@ionutivan @marsepu thank you. Let's centralize the discussion in #652 .

PR #652, opened 2016-12-06
reduce memory  usage, fix for issue #648 

marsepu:
1. Because work is done on a dispatch_queue and the queue does not guarantee when/if it will drain it's own autorelease pool objects are not released often and memory usage grows.

2. Retain cycles on self.

### DIFF
--- a/Analytics/Classes/Internal/SEGAnalyticsUtils.m
+++ b/Analytics/Classes/Internal/SEGAnalyticsUtils.m
@@ -43,12 +43,17 @@ BOOL seg_dispatch_is_on_specific_queue(dispatch_queue_t queue)
 void seg_dispatch_specific(dispatch_queue_t queue, dispatch_block_t block,
                            BOOL waitForCompletion)
 {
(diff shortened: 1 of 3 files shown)


tonyxiao:
@marsepu a few clarification questions

1) Wouldn't the autorelease pool eventually drain at the top of each event loop? 
2) I see the retain cycle. However because SEGSegmentIntegration is basically a singleton, why would the retain cycle matter as far as memory usage go?

tonyxiao:
So only (1) is the only thing that affects usage at the moment right? (2) is currently not causing any issues in production use?

marsepu:
YUP

f2prateek:
> (1) is the only thing that affects usage at the moment right? (2) is currently not causing any issues in production use?

In that case, could we split these changes up and address 1) and 2) in a seperate PR?

marsepu:
@f2prateek I set the files from the second commit back to how they were

tonyxiao:
LGTM. Thank you @marsepu

PR #655, opened 2016-12-29
Adding SEGBlockMiddleware helper and ability to add custom middleware into the middleware stack

tonyxiao:
This allows the creation of custom middlewares such as the following

https://github.com/segmentio/analytics-test-apps/blob/master/AnalyticsIOSTestApp/AnalyticsIOSTestApp/CustomMiddlewares.swift


f2prateek:
If I understand correctly, this PR is doing 2 main things:

1. Adds SEGMiddlewareBlock helper.
2. Allow adding custom middleware implementations.

Can we split this into 2 PRs?

This also should have tests, should be easy to do for 1) but we should verify for 2) as well.

tonyxiao:
Agreed on the tests. The changes feel small enough (and related enough, enabling the custom middlewares shown in the github above) to still be in the same PR though.

f2prateek:
I see - just felt logically independent since adding the block helper doesn't require the ability to set middlewares, nor does the ability to set middlewares depend on adding the block helper.

tonyxiao:
fair enough. I don't have strong preference here. Will add the tests and take it from here.

tonyxiao:
Added test. Codecov is happy. Everything passing and merging.

PR #656, opened 2017-01-03
Send build as string instead of integer.

f2prateek:
Ref: https://segment.atlassian.net/browse/BUGS-644

Previously the build was sent and stored as an integer. This can be incorrect since CFBundleVersion can be a string, e.g. `4.5.2`.

The change will check if there is a value stored under the old key. If
there is, it will copied over as a string to the new key, and the old key will be deleted.

This may send a truncated `previous_build` (i.e. "4" instead of "4.5.2") for
the first "Application Updated" call for customers migrating from an
affected version.

Tested locally. An event with the old SDK shows build as `1`: https://cloudup.com/cJY0mSyiLpj

An event with the new SDK shows build as `1.0.00` and previous_build as
`1` https://cloudup.com/cq5fLFpvBVr.

Note that because of this change, there will be an implicit "Application Updated" event triggered because the old Build and new build might not match. This might be not ideal for testing, but is ok in production because an app update will be required to roll this out anyway.

### DIFF
--- a/Analytics/Classes/SEGAnalytics.m
+++ b/Analytics/Classes/SEGAnalytics.m
@@ -94,7 +94,8 @@ static SEGAnalytics *__sharedInstance = nil;
 #pragma mark -
 
 NSString *const SEGVersionKey = @"SEGVersionKey";
-NSString *const SEGBuildKey = @"SEGBuildKey";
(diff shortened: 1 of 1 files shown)


tonyxiao:
👍

PR #664, opened 2017-01-31
Synchronize NSUserDefaults after we save values to it

tonyxiao:
To prevent race condition where app gets killed and changes do not save
(and thus Application Installed might fire twice)

cc @f2prateek


f2prateek:
LGTM

PR #667, opened 2017-02-06
fix: transaction.transactionIdentifier is nullable, letting it throug…

pgutierrezcf:
…h will crash the App in those cases where it is nil

According to Apple’s documentation (https://developer.apple.com/reference/storekit/skpaymenttransaction) the transactionIdentifier attribute is nullable. It is not clear under which circumstances a successful transaction will not have an identifier, but Apple’s overview of its APIs has determined that this is possible and we have tracked crashes linked to this condition in Crashlytics (http://crashes.to/s/8c3d076840e)

We cannot know if Segment would rather mark these transactions with a self-made identifiers or drop them, we assume that invalid transactions should not be tracked as they cannot be linked to records in other databases without this ID.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Analytics is available through [CocoaPods](http://cocoapods.org) and [Carthage](
 ### CocoaPods
 
 ```ruby
-pod "Analytics", "3.6.0-beta"
+pod "Analytics", "3.6.0-rc"
 ```
 
 If you'd rather have a version that lags behind but has been field tested longer, you may use


f2prateek:
@pgutierrezcf can you share the stack trace you are seeing without this change? The SDK should allow nil values through because of historical reasons.

pgutierrezcf:
Hello,

You can see the stack trace in the Crashlytics URL (http://crashes.to/s/8c3d076840e), or the attached file.

The problem with the nil value is that trackTransaction:forProduct: tries to create a NSDictionary with a literal, and inserting a nil value leads to a crash.

[CogniFit-App  29-01-2017, 14-00.crash.zip](https://github.com/segmentio/analytics-ios/files/756813/CogniFit-App.29-01-2017.14-00.crash.zip)

f2prateek:
@pgutierrezcf ah I see, that makes sense. I was incorrectly thinking that the crash  happens somewhere in the Analytics class/segment integration.

f2prateek:
@pgutierrezcf could you rebase your changes against `dev`? https://github.com/segmentio/analytics-ios/blob/master/CONTRIBUTING.md

pgutierrezcf:
It wasn't clear to me how to rebase the change in the context of a GitHub pull request, so I created a new pull request. Sorry if this represents an inconvenience.

https://github.com/segmentio/analytics-ios/pull/668

f2prateek:
no worries, thanks!

PR #672, opened 2017-02-28
Pass through userInfo

pinda:
Any reason you are not passing the userInfo that is received to the notification? But instead passing self which properties are mostly private and therefor not accessible.


f2prateek:
LGTM, cc @tonyxiao

tonyxiao:
LGTM as well. Circle failure looks transient. Rebuilding.

PR #674, opened 2017-03-07
Bug Fix #673

### LINKED ISSUE #673: Bug: SegAnalytics.shared.identify()
#647 introduced a bug where ```nil``` could no longer be passed into any of the `SegAnalytics.shared.identify(...)` methods in Swift. Previously, userId was allowed to have a value of ```nil```. This is no longer the case as it does not include the ```_Nullable``` clause. I'm currently opening a PR with this fix. If this update was intentional and ```nil``` is no longer able to be passed as a userId value then the comments should be updated to reflect that update.

chrisdhaan:
- Added ‘nullable’ and `_Nullable’ to SEGPayload userId property and
SEGAnalytics identify calls
- Fixes #673

- Ran `make build test` and received ** TEST SUCCEEDED ** in the console.

### DIFF
--- a/Analytics/Classes/Integrations/SEGIdentifyPayload.h
+++ b/Analytics/Classes/Integrations/SEGIdentifyPayload.h
@@ -5,7 +5,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface SEGIdentifyPayload : SEGPayload
 
-@property (nonatomic, readonly) NSString *userId;
+@property (nonatomic, readonly, nullable) NSString *userId;
 
 @property (nonatomic, readonly, nullable) NSString *anonymousId;
 
--- a/Analytics/Classes/SEGAnalytics.h
+++ b/Analytics/Classes/SEGAnalytics.h
@@ -65,9 +65,9 @@ NS_ASSUME_NONNULL_BEGIN
  When you learn more about who your user is, you can record that information with identify.
 
  */
-- (void)identify:(NSString *)userId traits:(SERIALIZABLE_DICT _Nullable)traits options:(SERIALIZABLE_DICT _Nullable)options;
-- (void)identify:(NSString *)userId traits:(SERIALIZABLE_DICT _Nullable)traits;
-- (void)identify:(NSString *)userId;
+- (void)identify:(NSString * _Nullable)userId traits:(SERIALIZABLE_DICT _Nullable)traits options:(SERIALIZABLE_DICT _Nullable)options;
+- (void)identify:(NSString * _Nullable)userId traits:(SERIALIZABLE_DICT _Nullable)traits;
+- (void)identify:(NSString * _Nullable)userId;
 
 
 /*!


f2prateek:
Does this only impact static analysis or have any impact at runtime as well? If it's the latter, is there a way we can add a test for this?

LGTM otherwise, cc @tonyxiao

chrisdhaan:
@f2prateek I believe the answer would be a runtime impact. I have a Swift project in which I had version 3.5.7 of the Analytics CocoaPod installed. Once I upgraded to 3.6.0 I could no longer run my application because I received an error for passing `nil` into the `identify(...)` method.

chrisdhaan:
@f2prateek I could add a test under the `persists userId` spec in the `AnalyticsTests.swift` class if you think that would be best. Let me know if I should add one.

f2prateek:
If there's a runtime impact, yup definitely let's add a test like:


```
 it("allows nil userId") {
     // Excercise a bug where nil could not be passed as the userId.
     // See https://github.com/segmentio/analytics-ios/pull/674.
      analytics.identify(nil, traits)

     ... 
     expect(analytics2.test_integrationsManager()?.test_segmentIntegration()?.test_userId()) == nil)
 }
```

I do think the method that doesn't accept traits should not allow a nil userId. That's an implementation bug if you're trying to call identify without either a userId or traits and the library should flag that.

tonyxiao:
@chrisdhaan what's the use case for having `nil` userId?

chrisdhaan:
@f2prateek that's probably correct. the method I was using was the method that accepts a userId, options, and traits as parameters. To confirm, you're purposing I update my changes to the following:

```objective-c
- (void)identify:(NSString * _Nullable)userId traits:(SERIALIZABLE_DICT _Nullable)traits options:(SERIALIZABLE_DICT _Nullable)options;
- (void)identify:(NSString *)userId traits:(SERIALIZABLE_DICT _Nullable)traits;
- (void)identify:(NSString *)userId;
```

where only

```objective-c
- (void)identify:(NSString * _Nullable)userId traits:(SERIALIZABLE_DICT _Nullable)traits options:(SERIALIZABLE_DICT _Nullable)options;
```

allows for a nullable userId?

chrisdhaan:
@tonyxiao the scenario is that I have not defined userId's for the applications users and am relying on the anonymous one generated by the Analytics framework. Previously, passing nil into the

```objective-c
- (void)identify:(NSString *)userId traits:(SERIALIZABLE_DICT _Nullable)traits options:(SERIALIZABLE_DICT _Nullable)options;
```

method was how I updated traits without having a userId. I would like to continue to do so unless the framework has updated to a point where userId is no longer optional. But if that's the case the documentation should reflect that. Does this answer your question?

f2prateek:
@chrisdhaan Mostly yep — think that `identify:nil traits:traits` and `identify:nil traits:traits options:options` should be supported (and calling `identify:nil` should be disallowed).

@tonyxiao This is used for updating traits for both an anonymous user (no userId is set) or an already known user (identify has been called with a userid before). This is also consistent with the spec, though maybe not communicated clearly which says an identify call is valid as long as either userId *or* anonymousId is provided https://segment.com/docs/spec/identify/.

I'm pretty sure *most* (if not all) our libraries allow this as this what's allowed by the spec. I did quick sanity check for a.js, android, java, python and node, all allow this.

tonyxiao:
Got it. @f2prateek i think we should make it more obvious (e.g. add an explicit api) to update traits without relying on passing `nil` for `userId`. It feels like a big hack right now and the intention is super unclear (should the userId be reset back to nil and user turned back into an anonymous visitor?)

tonyxiao:
@chrisdhaan what do you think of `identifyTraits:(_Nonnull NSDictionary *)traits`?

chrisdhaan:
@tonyxiao I do like what you're proposing and do believe it would make the intended functionality for identifying among the different methods more clear.

@f2prateek @tonyxiao is it still worth me making the updates suggested above that don't involve an explicit API addition to get this PR merged in as a temporary fix? I'm assuming if you were to add an explicit API that would require changing more code and also possibly having to update the other language frameworks which may take time. Let me know how to progress forward and I'll either push a commit or close this PR.

f2prateek:
@tonyxiao the android library has a method `identify:traits` for this and so does a.js, so agreed let's add it.

I also think that for consistency we should allow `nil` userId still and I don't think it's a "hack", where `identify:traits` is equivalent to `identify:nil traits:traits`. I think it's confusing because the behaviour isn't documented, but with documentation I don't think it would surprise people.

I haven't heard of people being confused with `analytics.identify(nil, traits)` was not "resetting" the userId.

tonyxiao:
Hmm it definitely would have confused me. Ok let's merge this PR for now and also add a separate api for `identify(traits)`

PR #675, opened 2017-03-07
Fix application opened

tonyxiao:
**What does this PR do?**
`Application Opened` is now sent when application moves from background to forground. 
`Application Opened` now includes properties listed in our mobile spec (`from_background`, `referring_application`,`url`) when application available.  For example, when launching app from a URL. 

https://segment.atlassian.net/browse/EPD-1153

**Questions:**
Application Backgrounded: https://github.com/segmentio/site-docs/pull/2158/
Application Opened was not updated in Mobile Spec as it is not consistent with Android functionality

@segmentio/gateway
cc @teresanesteby

### DIFF
--- a/Analytics/Classes/SEGAnalytics.m
+++ b/Analytics/Classes/SEGAnalytics.m
@@ -42,21 +42,21 @@ static SEGAnalytics *__sharedInstance = nil;
 - (instancetype)initWithConfiguration:(SEGAnalyticsConfiguration *)configuration
 {
     NSCParameterAssert(configuration != nil);
-    
+
     if (self = [self init]) {
         self.configuration = configuration;
         self.enabled = YES;
-        
+
         // In swift this would not have been OK... But hey.. It's objc
         // TODO: Figure out if this is really the best way to do things here.
         self.integrationsManager = [[SEGIntegrationsManager alloc] initWithAnalytics:self];
         
         self.runner = [[SEGMiddlewareRunner alloc] initWithMiddlewares:
                        [configuration.middlewares ?: @[] arrayByAddingObject:self.integrationsManager]];
-        
+
         // Attach to application state change hooks
         NSNotificationCenter *nc = [NSNotificationCenter defaultCenter];
-        
+
         // Pass through for application state change events
         for (NSString *name in @[ UIApplicationDidEnterBackgroundNotification,
                                   UIApplicationDidFinishLaunchingNotification,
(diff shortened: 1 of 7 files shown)


f2prateek:
1. Can we start adding tests for these?
2. The original behaviour was put in place so that even if you don't initialize segment in `didFinishLaunch` phase, we could track an 'App Opened' event. Even though that's what's recommended, people don't always follow it. I agree that this is the right behaviour, so we should document this for `trackApplicationLifecycleEvents` setter and in site-docs.

tonyxiao:
@f2prateek do you have a good pattern for testing these hooks? And yes agreed we should update the docs @TeresaNesteby

TeresaNesteby:
@f2prateek @tonyxiao I am working on the ios lib docs update and will link here when I have a PR.

f2prateek:
@tonyxiao with the middleware change, I think we should be able to have a "test" middleware that we can verify is invoked with this event.

tonyxiao:
True. Will work on that with @TeresaNesteby on thursday.

Analytics/Classes/SEGAnalytics.m
```diff
@@ -99,13 +97,24 @@ - (void)dealloc
...
+        [self _applicationWillEnterForeground];
     }
+}
 
-    // Previously SEGBuildKey was stored an integer. This was incorrect because the CFBundleVersion
+- (void)_applicationDidFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+    if (!self.configuration.trackApplicationLifecycleEvents) {
+        return;
+    }
+        // Previously SEGBuildKey was stored an integer. This was incorrect because the CFBundleVersion
```
f2prateek: I think the formatting is off here. I just run https://github.com/square/spacecommander to clean it up.

tonyxiao: @TeresaNesteby wanna take a stab at this? :)

tonyxiao:
@f2prateek how come closed?

f2prateek:
@tonyxiao Sorry, I accidentally deleted the branch as I was pruning old ones. Restored it but forgot to reopen.

Analytics/Classes/SEGAnalytics.m
TeresaNesteby: @f2prateek @tonyxiao Just pushed up updates to indentation. I used Xcode's default to reindent.  Let me know if it doesn't match up with what you are seeing with spacecommander.

Analytics/Classes/SEGAnalytics.m
```diff
@@ -164,10 +164,10 @@ - (void)_applicationWillEnterForeground {
     NSString *currentVersion = [[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"];
     NSString *currentBuild = [[NSBundle mainBundle] infoDictionary][@"CFBundleVersion"];
     [self track:@"Application Opened" properties:@{
-        @"from_background": @YES,
-        @"version" : currentVersion,
-        @"build" : currentBuild,
-    }];
+                                                   @"from_background": @YES,
```
tonyxiao: @TeresaNesteby did you re-format this by hand or use a tool? New indentation seems off actually

TeresaNesteby: @tonyxiao I used Xcode - structure - re-indent. I'll try using spacecommander - that prateek mentioned.

Analytics/Classes/SEGAnalytics.m
f2prateek: You can use this via xcode https://github.com/travisjeffery/ClangFormat-Xcode. Iirc it has the same output as space commander.

tonyxiao:
fyi @TeresaNesteby @f2prateek

f2prateek:
LGTM!

PR #677, opened 2017-03-26
Update CHANGELOG.md

TeresaNesteby:
**What does this PR do?**
It adds warning flag in changelog - users should not be on version 3.5.0 or 3.5.1. 

The bug introduced in 3.5.0 was fixed in version 3.5.2


f2prateek:
In general, any older release that is followed by bugfixes in newer releases is not recommended.

PR #679, opened 2017-03-28
Add [session finishTasksAndInvalidate] to SEGHTTPClient.m to prevent memory leak

ethanhuang13:
**What does this PR do?**

Add` [session finishTasksAndInvalidate]` to SEGHTTPClient.m to prevent memory leak

You should release the `NSURLSession` object after using it. Details:

Segment SDK is using `[NSURLSession sessionWithConfiguration:config]` to create NSURLSession object for each task rather than a singleton object.

In Apple's [document](https://developer.apple.com/reference/foundation/urlsession), it says: 

> **Important**
> The session object keeps a strong reference to the delegate until your app exits or explicitly invalidates the session. If you do not invalidate the session, your app leaks memory until it exits.

There is also an StackOverflow question discussed about this: [http://stackoverflow.com/a/35757989/2627067](http://stackoverflow.com/a/35757989/2627067)

When using Xcode 8's memory debugging tool, you will see a lot of memory leak caused by Segment SDK. 

You should add `[session finishTasksAndInvalidate]` after `[task resume]` to allow outstanding tasks to finish before invalidating the object.

Use Xcode 8's memory debugging tool, compare the result with or without `[session finishTasksAndInvalidate]`.

**Questions:**

@segmentio/gateway

### DIFF
--- a/Analytics/Classes/Internal/SEGHTTPClient.m
+++ b/Analytics/Classes/Internal/SEGHTTPClient.m
@@ -100,6 +100,7 @@
         completionHandler(YES);
     }];
     [task resume];
+    [session finishTasksAndInvalidate];
     return task;
 }
 
@@ -140,6 +141,7 @@
         completionHandler(YES, responseJson);
     }];
     [task resume];
+    [session finishTasksAndInvalidate];
     return task;
 }
 
@@ -200,6 +202,7 @@
         completionHandler(YES, responseJson);
     }];
     [task resume];
+    [session finishTasksAndInvalidate];
     return task;
 }
 


tonyxiao:
@ethanhuang13 reading through the stack overflow post seems that the right solution is actually to use a singleton session rather than initialize one per request? Also re: delegate, we aren't actually using session delegates at the moment, so there isn't an retain cycle there. Is the session retained globally by Apple until `invalidate` is called somehow?

ethanhuang13:
The retain cycle is kept in the session object internally. If you create a new session object other than `sharedSession`, in order to release it, you should call `finishTasksAndInvalidate` or `invalidateAndCancel`. See the official doc I linked above. I think its clear.

There are 2 right solutions. One is using singleton, the other is call `[session finishTasksAndInvalidate]` after `[task resume]`

For example, Facebook and Mixpanel use singleton, while [Branch.io](https://github.com/BranchMetrics/ios-branch-deep-linking/search?utf8=✓&q=finishTasksAndInvalidate&type=) use `finishTasksAndInvalidate`. They are both valid solutions.

tonyxiao:
@ethanhuang13 the docs you linked to says

> you allocate new memory for a new session with each request, then you will see unbounded memory growth from SSL caches that persist for 10 minutes regardless of whether or not you invalidateAndCancel or flush/reset the session

Which leads me to believe that only the singleton solution is the best one?

ethanhuang13:
My code is to patch the memory leak of session objects upon current SDK code structure. 

If you want to save the memory usage of caching SSL mentioned in the SO post, you might want to add a session property in the SDK singleton. (I think that will be another PR. )

tonyxiao:
Fair enough, will merge this for now and have separate ticket to switch to singleton.

min:
@tonyxiao we've been observing this leak as well. when will a new version of the SDK with this fix be released?

tonyxiao:
@min new release will be out within the next day.

tonyxiao:
@min 3.6.1 has just been released.

PR #682, opened 2017-04-04
Add Builder types

f2prateek:
**What does this PR do?**
Add builder types for payloads. This makes it easier to create payload instances.

Not required. Just verify the automated tests are correct.

Eventually this will be used to simplify how the complicated middlewares are implemented. But not just yet.

@segmentio/gateway

### DIFF
--- a/Analytics/Classes/Integrations/SEGAliasPayload.m
+++ b/Analytics/Classes/Integrations/SEGAliasPayload.m
@@ -1,6 +1,18 @@
 #import "SEGAliasPayload.h"
 
 
+@implementation SEGAliasPayloadBuilder
+
+- (instancetype)init
+{
+    if (self = [super init]) {
+    }
+    return self;
+}
+
+@end
+
+
 @implementation SEGAliasPayload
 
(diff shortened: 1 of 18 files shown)


Analytics/Classes/Integrations/SEGAliasPayload.m
```diff
@@ -13,4 +25,17 @@ - (instancetype)initWithNewId:(NSString *)newId
     return self;
 }
 
+- (instancetype)initWithBuilder:(SEGAliasPayloadBuilder *)builder
+{
+    if (self = [super initWithBuilder:builder]) {
+        NSParameterAssert(builder);
+
+        NSString *theNewId = [builder.theNewId copy];
```
stephenmathieson: nit: prefixing variable names with "the" seems odd :P

f2prateek: Try taking it out :)

It's an objective c thing. You can't use new as a prefix 
http://stackoverflow.com/questions/24308162/property-name-starting-with-new-prefix-leads-to-bad-access-error-in-ios

Issue #683, opened 2017-04-18
Deadlock issue with forwardSelector:arguments:options: method


bbernberg:
We are seeing a repeatable deadlock issue due to the `@synchronized(self)` lock in `SEGIntegrationsManager.m`'s `forwardSelector:arguments:options:` method.  We are seeing the issue when the app becomes active and the `handleAppStateNotification` path is executed.

bbernberg:
FYI, here is an example snapshot of the deadlock.  As can be seen, Thread 26 (which has the lock) is attempting to access the main thread, which is attempting to obtain lock:

![screen shot 2017-04-18 at 5 52 13 pm](https://cloud.githubusercontent.com/assets/1385086/25154746/ded29d80-245f-11e7-841e-f0fe307689a9.png)

raylillywhite:
Pretty sure this is the same issue, but the stack trace where we're seeing this is a bit different, so in case it's helpful for debugging:

![](http://d.pr/i/tBz5/kzlUFiG6+)
![](http://d.pr/i/13S3J/5sPKOngV+)

f2prateek:
Thanks for the report! @tonyxiao is looking into this.

bbernberg:
FYI I tried a different approach using serial dispatch queues & a serial `NSOperationQueue` but both approaches ran into the same problem.  Is there a reason the App State Notifications need to be handled synchronously from the main thread?  I don't have enough background knowledge of the full system to understand if this is truly a requirement.

tonyxiao:
@bbernberg I'm trying to investigate this right now. Do you happen to have a minimal example that you can put on github reproducing this issue?

raylillywhite:
This regularly reproduces it on my iPhone 7 Plus from backgrounding and foregrounding the app a couple times: https://gist.github.com/raylillywhite/cb74d4d1b9325bfcde94119170be55ec

In the app where I'm experiencing it, we're obviously not sending anywhere near that many events, and it happens every time we turn on airplane mode, launch the app, background it, turn off airplane mode, foreground the app. I wasn't able to successfully reproduce that with a minimal example though.

bbernberg:
@raylillywhite 's example looks perfect.  It definitely seems to be the App State Notifications deadlocking with other events fired from the main thread.

tonyxiao:
[...]

PR #684, opened 2017-04-25
Use a separate queue for endBackgroundTask to fix deadlock

tonyxiao:
Hotfix for https://github.com/segmentio/analytics-ios/issues/683

We can sometimes deadlock on `@synchronized(self)` block inside `SEGIntegrationsManager` because both the main queue and `io.segment.analytics.segmentio` queue are trying to acquire a single synchronized lock while the main thread waits for `io.segment.analytics.segmentio` queue via `dispatchBackgroundAndWait` during `endBackgroundTask`. This PR uses a separate queue for `endBackgroundTask` to avoid that issue. This works because the only thing we are trying to make sure is access to `self.backgroundTaskId` is thread-safe. There are no other dependencies on other operations sent to  `io.segment.analytics.segmentio` queue

cc @f2prateek


f2prateek:
LGTM, would love to get a test case up as we discussed.

tonyxiao:
Will add tests separately.

Issue #670, opened 2017-02-23
Calling reset method cancels previous track event request.


aleksandergrzyb:
**What's the problem:**

Tracking some event just before calling `reset` method cancels the request resulting in with no event being tracked:

**Detailed description:**

In [this line](https://github.com/segmentio/analytics-ios/blob/master/Analytics/Classes/Internal/SEGSegmentIntegration.m#L515) there is a cancellation of `self.batchRequest`. If the request is under progress the tracking event is never send into Segments. 

```swift
SEGAnalytics.shared().track(...) // Never tracked
SEGAnalytics.shared().reset() // Cancels previous tracking event
```

**How to reproduce:**

1. Turn on debugging via `SEGAnalytics.debug(true)`.
2. Track some event and on the next line call immediately `reset()` method.
3. Logs show that track event was cancelled.

**How to fix?**

To get around the issue we are using `dispatch_after` where we call `reset`. Do you have any other ideas how it could be resolved? Is it necessary to cancel the request inside of `reset`?

tonyxiao:
@f2prateek what do you think is the right behavior here? Also what happens on analytics-android?

tonyxiao:
Related issue https://github.com/segmentio/analytics-ios/issues/681

f2prateek:
Reset is meant for the user logging out case, so I think the behavior should be to keep the existing events and flush them later. On Android we don't clear the queue.

aleksandergrzyb:
Currently in production we have `flushAt` property set to `5`. When user logs out we just want to clear all user information, but we want to send previously generated events by this user. Right now we maintain [a fork](https://github.com/aleksandergrzyb/analytics-ios/commit/ae958edefc972f6510ea081976f5e24743d40433) where we implemented that for ourselves.

If you agree with it I am happy to create PR to official repository.

tonyxiao:
I like what you have in the fork. Can you create a PR based on it? Thanks you @aleksandergrzyb

aleksandergrzyb:
Yes, will do tomorrow. 👍

aleksandergrzyb:
Created PR #690.

PR #690, opened 2017-05-10
Add method for clearing user data

aleksandergrzyb:
**What does this PR do?**

Adds a method to clear user data without canceling ongoing requests. See #670.

In `SEGSegmentIntegration` file where the details of the implementation are.

Set `flushAt` to some high value, maybe 30. Generate tracking events and then call `clearUserData`. All generated tracking events should be send.

We were calling `reset` after user logged out in our application, but we quickly noticed that calling `reset` also cancels any ongoing request. Since we had `flushAt` set to 5 some of the request generated by the user just before log out action were not send to Segment integration.

#670 

@segmentio/gateway

### DIFF
--- a/Analytics/Classes/Integrations/SEGIntegration.h
+++ b/Analytics/Classes/Integrations/SEGIntegration.h
@@ -45,8 +45,11 @@ NS_ASSUME_NONNULL_BEGIN
 // @see https://segment.com/docs/spec/alias/
 - (void)alias:(SEGAliasPayload *)payload;
 
-// Reset is invoked when the user logs out, and any data saved about the user should be cleared.
+// Reset is invoked when the user logs out, and any data saved about the user and ongoing or cached requests should be cleared.
 - (void)reset;
+
+// Clear user data is invoked when the user logs out, and any data saved about the user should be cleared.
+- (void)clearUserData;
 
 // Flush is invoked when any queued events should be uploaded.
 - (void)flush;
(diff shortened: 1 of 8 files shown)


f2prateek:
Rather than add a new method, why not have reset not clear in flight requests or the queue?

The reason why I'd prefer not to add a new method is that we'll now have to go through and update other our client side integrations and and make sure they map the new `clearUserData` method. I don't think that's worth the effort as most tools I've worked with have a single method for clearing the local data (where they don't delete queued events) - they don't have separate methods for `reset` and `clearUserData`.

ladanazita:
Backing up @f2prateek  here. Creating a new method will lead to a ton of work to ensure this behavior is applied across all integrations. 

Can we improve the existing `reset` behavior or are there implications that this would cause a breaking change?

aleksandergrzyb:
I didn't want to modify original `reset` method, because if someone relies on canceling ongoing requests that would be a breaking change, so I created an additive one. @tonyxiao confirmed in #670 based on implementation [in our fork](https://github.com/aleksandergrzyb/analytics-ios/commit/ae958edefc972f6510ea081976f5e24743d40433) that this is a good solution for the problem that we are experiencing.

If `reset` method can be modified to clear only user data and traits I am happy to change the PR and submit respective changes. Let me know what you think, and thanks! 👍

f2prateek:
There could be arguments made either way, but I don't think it'd be a breaking API change. I don't think `reset` method defines it as a contract. It's a side effect of the existing implementation, but certainly nothing in the docs suggest that it is expected to cancel outgoing requests or clear the queue.

FWIW that's also the behaviour of analytics.js and analytics-android.

aleksandergrzyb:
Also are you sure this won't break nothing internally?

So the change finally could look like [this](https://gist.github.com/aleksandergrzyb/33da037ed4a4891db8bcc822b307f1da) (+ test). Please confirm if it's ok with you. If yes, I will make the changes.

f2prateek:
> Also are you sure this won't break nothing internally?

Can you clarify what you mean? I don't think anything would break as that's the exact implementation you're suggesting for the `clearUserData` method.

aleksandergrzyb:
I created a new PR #691 where I implemented what we discussed here. Closing this PR.

PR #691, opened 2017-05-17
Remove canceling ongoing requests in reset method.

aleksandergrzyb:
**What does this PR do?**

Removes calls to cancel ongoing requests. See #670.

In `SEGSegmentIntegration.m` file where the calls where removed.

Set `flushAt` to some high value, maybe 30. Generate tracking events and then call `reset`. All generated tracking events should be send.

We were calling `reset` after user logged out in our application, but we quickly noticed that calling `reset` also cancels any ongoing request. Since we had `flushAt` set to 5 some of the request generated by the user just before log out action were not send to Segment integration.

#670 

@segmentio/gateway

### DIFF
--- a/Analytics/Classes/Internal/SEGSegmentIntegration.m
+++ b/Analytics/Classes/Internal/SEGSegmentIntegration.m
@@ -513,18 +513,13 @@ static CTTelephonyNetworkInfo *_telephonyNetworkInfo;
         [self.storage removeKey:SEGUserIdKey];
 #if TARGET_OS_TV
         [self.storage removeKey:SEGTraitsKey];
-        [self.storage removeKey:SEGQueueKey];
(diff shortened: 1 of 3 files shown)


aleksandergrzyb:
@f2prateek the changes proposed here it's what you had in mind?

f2prateek:
LGTM, cc @tonyxiao

tonyxiao:
@f2prateek didn't we want a separate called `clearUserData` for this usecase?

f2prateek:
No we don't https://github.com/segmentio/analytics-ios/pull/690#issuecomment-300579079.

Maybe eventually, but not until we can properly migrate integrations.

tonyxiao:
ok LGTM then.

PR #696, opened 2017-05-26
Add missing includes to umbrella header

msanders:
Fixes 
screenshot 2017-05-26 10 33 28

when importing Segment via Carthage.

### DIFF
--- a/Analytics/Analytics.h
+++ b/Analytics/Analytics.h
@@ -17,3 +17,5 @@ FOUNDATION_EXPORT const unsigned char AnalyticsVersionString[];
 #import "SEGAnalytics.h"
 #import "SEGSegmentIntegration.h"
 #import "SEGSegmentIntegrationFactory.h"
+#import "SEGContext.h"
+#import "SEGMiddleware.h"


bbeaird:
@f2prateek - Can we get this merged? Another biz tier customer mentioned this today.

f2prateek:
LGTM, @tonyxiao was this an oversight or intentional?

bbeaird:
FYI - from a biz tier customer... 

```
I'd recommend your ios devs take a more holistic view at this. There is an interesting/confusing mix of public/project/private header tagging among all of the folders in your project, even in the Internal folder which one might think would not be public. In addition, as I was rolling back some of these dependencies, I suspected that some of this public sprawl probably occurred in an attempt to solely make SEGIntegration available for unit tests. That problem could probably be better solved using a different approach.
```

benshan:
fwiw, this PR is a nice simple quick win. I have applied it locally to work around build errors when importing SEGAnalytics into a swift file. The more robust fix to this could be done as a followup PR, imo.

tonyxiao:
@msanders thank you for the PR. And yes @benshan I think a more comprehensive review of what should be `public`, `private` and `project` visibility is definitely in order. Will add it to the roadmap.

Issue #213, opened 2014-10-15
"officially" support iOS 8 extensions


piemonte:
hey there,

iOS 8 enables the ability to develop out of app extensions. the architecture of these extensions make some api unavailable, such as UIApplication.

it would be nice to define a means for avoiding it's use in the future or just updating the tutorial to explicitly state the install is different.

here was the issue tracking AFNetworking changes:

AFNetworking/AFNetworking#2119

robmaceachern:
+1

Is there a workaround available right now?

pomarec:
+1

hunghoangterralogic:
Hey, is there any update on this ?

f2prateek:
For now, the workaround is to use our HTTP API directly https://segment.com/docs/sources/server/http/

tvarady:
Is this on your roadmap? Or should we start to think about workarounds?

f2prateek:
It's still on the roadmap (the issue will be closed if it is not). No timeline yet, as mentioned you can use the HTTP API directly for now.

danieljimenez:
That's not really a work around for apps that expect to be able to handle events while not connected to the internet.

f2prateek:
@danieljimenez yes you will have to write your own queueing mechanisms. If you need a starting point, you can fork the SDK and remove the bits you don't need.

tvarady:
Now I checked Segment's HTTP API, and it looks like I have to re-implement parts of Segment iOS SDK to make it work... I have to collect Context data, identify user, implement NSURLSession to access the API, just like the SDK itself does. So it's not so trivial, and it's just because some lines are not compiling in extension context.

f2prateek:
@tvarady you could fork the sdk in the meantime and remove/change the bits you don't need, likely you don't need to write one from scratch. If you do, let us know and that will help us bring that functionality into the core library!

LeoNatan:
It's embarrassing that an issue like this remains open since 2014. Shows how serious the developers are about iOS development. This should turn a warning sign for any iOS developer looking to use this framework or service. Simply do not.

f2prateek:
@LeoNatan I apologize for the delay. We definitely want to support extensions (which is why the issue is still open). It hasn't received as much attention because it is one of our least requested features — there are <15 people combined who have expressed interest on this issue and via our internal Zendesk so far 

[...]

PR #698, opened 2017-06-13
Extract UIApplication to permit linkage with iOS extensions.

benshan:
**What does this PR do?**
Extract usage of UIApplication into protocol and optional property on config, so the sdk can be used with iOS extensions.

SEGAnalyticsConfiguration.h

Link with an iOS app, set breakpoints in the appropriate UIApplication notification handlers.

Also set APPLICATION_EXTENSION_API_ONLY=YES to guard this at build time.

#213 

NA

**Questions:**
Yes, now by default the optional config.application will be nil unless a non-extension target sets it to UIApplication.shared.

No

Not sure. Would be nice to build out TestApplication more and add a test to fire UIApplicationDidEnterBackground, and assert that a bg task is added, then removed when the handler is called.

@segmentio/gateway

### DIFF
--- a/Analytics.xcodeproj/project.pbxproj
+++ b/Analytics.xcodeproj/project.pbxproj
@@ -722,6 +722,7 @@
 		EADEB8641DECD080005322DA /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				APPLICATION_EXTENSION_API_ONLY = YES;
 				CODE_SIGN_IDENTITY = "";
 				DEFINES_MODULE = YES;
 				DYLIB_COMPATIBILITY_VERSION = 1;
@@ -743,6 +744,7 @@
 		EADEB8651DECD080005322DA /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				APPLICATION_EXTENSION_API_ONLY = YES;
 				CODE_SIGN_IDENTITY = "";
 				DEFINES_MODULE = YES;
 				DYLIB_COMPATIBILITY_VERSION = 1;
--- a/Analytics/Classes/Integrations/SEGIntegrationsManager.m
+++ b/Analytics/Classes/Integrations/SEGIntegrationsManager.m
@@ -90,7 +90,10 @@ static NSString *const kSEGAnonymousIdFilename = @"segment.anonymousId";
         NSNotificationCenter *nc = [NSNotificationCenter defaultCenter];
 
         // Update settings on foreground
-        [nc addObserver:self selector:@selector(onAppForeground:) name:UIApplicationWillEnterForegroundNotification object:nil];
+        id<SEGApplicationProtocol> application = configuration.application;
+        if (application) {
+            [nc addObserver:self selector:@selector(onAppForeground:) name:UIApplicationWillEnterForegroundNotification object:application];
+        }
     }
     return self;
 }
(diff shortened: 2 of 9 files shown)


tonyxiao:
This is a super comprehensive PR. Thank you very much for this. My only question would be what happens when the library is used from an iOS application? Wouldn't user now need to explicitly set `[UIApplication sharedApplication]` to `SEGAnalyticsConfiguration` otherwise library will behave in backward incompatible ways? Can we configure it automatically based on whether or not `UIApplication` is available at runtime?

benshan:
@tonyxiao, yes, that was my main concern as well. The best I can come up with for runtime config is to use objc runtime to try to get sharedApplication via string name. That should appease the compiler while still getting the benefits of the build setting to warn of regression. (even better if all the warnings are cleaned and then enable treat warnings as errors)  We should probably test this approach both in an app target and in an extension to see if it works properly.

benshan:
@tonyxiao see last commit that I pushed. This is what I'm thinking. It seems to work in my testing, and it doesn't interfere with APPLICATION_EXTENSION_API_ONLY = YES.

benshan:
If you think a code comment is necessary in that new block to explain why it's done that way, let me know.

benshan:
Also note that you could use the technique in the last commit with the original code, without extraction of UIApplication protocol and optional injection. I think the DI is nicer because we can now unit test these flows, but it's really up to your team wrt what you prefer.

tonyxiao:
This is awesome! Thank you for the thoughtfulness and fine attention here. Merging.

pixelmatrix:
FYI this PR is causing some issues in our Swift application. We use background tasks for various operations, and after upgrading to this version the compiler complains about `UIApplication.shared.endBackgroundTask(_)` with a UIBackgroundTaskIdentifier (Int). It now wants a UInt instead. It's easy enough to convert, but feels wrong to need to do that.

benshan:
@pixelmatrix what is your SDKROOT? At least with iOS10, Apple has these background task APIs using a type `UIBackgroundTaskIdentifier` which is `typedef NSUInteger UIBackgroundTaskIdentifier;` from UIApplication.h. So likely you would need to change this to avoid warnings in the future.

Our direct conformance cover might be making that more front and center, though I haven't seen this issue in our codebase which is a mix of objc/swift. An alternative for this PR is to change `SEGApplicationProtocol` to be a protocol which does not use the Apple types, then wrap/marshal in the conformance extension.

benshan:
Depending on all the versions of iOS that Segment supports, perhaps the easier thing to do is add a conditional compile here for the type, then include the different variations Apple has applied over those versions. I don't have UIApplication.h handy from earlier versions, but I assume this API directly used NSInteger at one point and now has been extracted as a typedef and the underlying type changed.

pixelmatrix:
Our SDK is set to 10.3, so it's using the most recent version. I'm seeing NSUInteger in UIApplication.h, but looking at the swift headers I see `public typealias UIBackgroundTaskIdentifier = Int`

benshan:
@pixelmatrix I am able to repro, and I coded a decent workaround, so please don't pollute your code with the conversion. Give me a day or so to clean it up and submit here. I tested with Xcode9b3 and this is no longer an issue there, so likely was a stale manual Swift mapping in UIApplication which got out of sync.

pixelmatrix:
Awesome, thank you for the prompt response

benshan:
@pixelmatrix see referenced PR. Can you try applying locally and confirming that it fixes for you? (I have added an automation test as well, TDD style, which would have caught it.)

pixelmatrix:
@benshan just applied that branch to our app and it compiled just fine. looks like it's fixed!

benshan:
Thanks @pixelmatrix for confirming!

Issue #692, opened 2017-05-23
Improper use of NSURLSession creation.


thefaj:
SEGHTTPClient.m improperly generates a new NSURLSession each time data is uploaded to Segment's API. Apple's recommendation on this since iOS 7 has been to generate your NSURLTasks on a limited number of reusable NSURLSessions.

Segment's current implementation causes memory leaks when measured in Instruments.

f2prateek:
This seems reasonable. Our assumption was that most apps don't upload data more than once or twice per session — so it's not the end of the world if we create 1-2 new sessions. But does make sense people might have lower thresholds in production too, so we should guard it. cc @tonyxiao

jdmaresco:
We're seeing memory leaks – and many crashes on older devices. We've isolated the issue to our iOS app, and the Segment SDK. It seems that a new NSURLSession is created for every event request, and then never closed. Many users stay in our app for 2-3 minutes and fire 20-30 events. 

Is there a priority fix going in for this?

f2prateek:
Are you using a low batch/flush size or interval? That might explain why many sessions are being created, though not why there aren't being closed (we do this here https://github.com/segmentio/analytics-ios/blob/dev/Analytics/Classes/Internal/SEGHTTPClient.m#L103). Additionally we only create a single session at a time.

cc @tonyxiao on status of the fix.

thefaj:
That's not correct @f2prateek. Your library creates a new session with every network request. This is incorrect usage of NSURLSession and causes leaks.

This plus the major security issues in the public sdk makes me question how you guard our users' extremely sensitive data inside your company. Segment.io is not safe for use.

f2prateek:
I think you were confused by the statement "Additionally we only create a single session at a time." - what I meant to clarify there was we only make one network request at a time and hence create a single session at a time - I do agree we create multiple sessions (one per request) sequentially.

jdmaresco:
@f2prateek Yes we have batch size to the minimum as we want it to be fairly real-time. I don't believe this should be an issue at all if NSURLSessions are used (and closed) correctly.

jdmaresco:
@f2prateek Do you have an update on this?

f2prateek:
@jdmaresco, @tonyxiao will let you know

tonyxiao:
https://github.com/segmentio/analytics-ios/pull/699 @jdmaresco @f2prateek

f2prateek:
Closed by #699

[...]

PR #699, opened 2017-06-16
Reuse NSURLSession in SEGHTTPCLient

tonyxiao:
Solves issue https://github.com/segmentio/analytics-ios/issues/692#issuecomment-307870908

Reuse the session in client so we do not create it every time. 

When client deallocs, we invalidate the session.

### DIFF
--- a/Analytics/Classes/Internal/SEGHTTPClient.h
+++ b/Analytics/Classes/Internal/SEGHTTPClient.h
@@ -15,6 +15,8 @@ NS_ASSUME_NONNULL_BEGIN
 @interface SEGHTTPClient : NSObject
 
 @property (nonatomic, strong) SEGRequestFactory requestFactory;
+@property (nonatomic, readonly) NSMutableDictionary<NSString *, NSURLSession *> *sessionsByWriteKey;
+@property (nonatomic, readonly) NSURLSession *genericSession;
 
 + (SEGRequestFactory)defaultRequestFactory;
 + (NSString *)authorizationHeader:(NSString *)writeKey;
--- a/Analytics/Classes/Internal/SEGHTTPClient.m
+++ b/Analytics/Classes/Internal/SEGHTTPClient.m
@@ -28,22 +28,42 @@
         } else {
             self.requestFactory = requestFactory;
         }
+        _sessionByWriteKey = [NSMutableDictionary dictionary];
+        NSURLSessionConfiguration *config = [NSURLSessionConfiguration defaultSessionConfiguration];
+        config.HTTPAdditionalHeaders = @{@"Accept-Encoding" : @"gzip"};
+        _genericSession = [NSURLSession sessionWithConfiguration:config];
     }
     return self;
+}
+
+- (NSURLSession *)sessionForWriteKey:(NSString *)writeKey {
+    NSURLSession *session = self.sessionsByWriteKey[writeKey];
+    if (!session) {
+        NSURLSessionConfiguration *config = [NSURLSessionConfiguration defaultSessionConfiguration];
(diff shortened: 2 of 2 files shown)


Analytics/Classes/Internal/SEGHTTPClient.h
```diff
@@ -15,6 +15,8 @@ NS_ASSUME_NONNULL_BEGIN
 @interface SEGHTTPClient : NSObject
 
 @property (nonatomic, strong) SEGRequestFactory requestFactory;
+@property (nonatomic, readonly) NSMutableDictionary<NSString *, NSURLSession *> *sessionsByWriteKey;
```
f2prateek: There's only ever one writeKey (the client must be a singleton), so this will only have one entry? In which case do we need a map>

Analytics/Classes/Internal/SEGHTTPClient.m
```diff
@@ -28,22 +28,42 @@ - (instancetype)initWithRequestFactory:(SEGRequestFactory)requestFactory
         } else {
             self.requestFactory = requestFactory;
         }
+        _sessionByWriteKey = [NSMutableDictionary dictionary];
+        NSURLSessionConfiguration *config = [NSURLSessionConfiguration defaultSessionConfiguration];
+        config.HTTPAdditionalHeaders = @{@"Accept-Encoding" : @"gzip"};
```
f2prateek: I think the formatter didn't run and left `@{@` insead of correcting to `@{ @`

jdmaresco:
@f2prateek @tonyxiao do you guys plan to fix and merge this in soon? Based on our research, we are pretty sure the memory implications of this change will help us reduce crashes for up to 5% of our users. Eager to see a resolution here. cc: @volonbolon

Analytics/Classes/Internal/SEGHTTPClient.h
tonyxiao: While that's true the interface of the SEGHTTPClient is such that we cannot guarantee that only one writeKey is used. I didn't want to make extensive modifications so this is where we ended up.

Analytics/Classes/Internal/SEGHTTPClient.m
tonyxiao: 👍

Analytics/Classes/Internal/SEGHTTPClient.h
f2prateek: The SEGHTTPClient isn't a singleton, so it'll only ever be tied to a single SEGAnalytics instance, and hence only ever talk to a single writeKey?

tonyxiao: Agreed that's true in practice, but from the isolated perspective of SEGHTTPClient, it cannot guarantee how it will be used. What's the downside of the current approach anyway?

f2prateek: Just seems like an unnecessary extra step. Even if the http client was being used by multiple writeKeys, why wouldn't you want to reuse the sessions between the writeKeys?

tonyxiao:
@jdmaresco will be released next monday.

volonbolon:
Code has been merged to "dev", not master. How stable is dev?

volonbolon:
@tonyxiao please advice. Is dev stable enough to be used in production code?

tonyxiao:
@volonbolon we will be releasing a new version today with changes from dev that have been baking for a little while.

tonyxiao:
https://github.com/segmentio/analytics-ios/releases/tag/3.6.2 released

PR #702, opened 2017-07-07
Fix NSURLSession being prematurely invalidated

tonyxiao:
These two separate PRs unfortunately introduced conflicting changes

https://github.com/segmentio/analytics-ios/pull/699
https://github.com/segmentio/analytics-ios/pull/679

When we re-use session, we should no longer invalidate session right after each task.


tonyxiao:
Tested manually and working

PR #711, opened 2017-08-24
Use DEBUG preprocessor flag to conditionally disable assertions in prod

tonyxiao:
@ladanazita @TeresaNesteby 

Tested manually w/ @TeresaNesteby in test app and confirmed working.

### DIFF
--- a/Analytics/Classes/Internal/SEGAnalyticsUtils.m
+++ b/Analytics/Classes/Internal/SEGAnalyticsUtils.m
@@ -146,6 +146,7 @@ static id SEGCoerceJSONObject(id obj)
 
 static void AssertDictionaryTypes(id dict)
 {
+#ifdef DEBUG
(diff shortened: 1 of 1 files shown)


Analytics/Classes/Internal/SEGAnalyticsUtils.m
```diff
@@ -146,6 +146,7 @@ static id SEGCoerceJSONObject(id obj)
 
 static void AssertDictionaryTypes(id dict)
 {
+#ifdef DEBUG
```
NachoSoto: This should be `#if DEBUG`.
If DEBUG is defined as `0` this would evaluate as true :)

tonyxiao: Good point. However in practice that's not how most projects (including `Pods.xcodeproj`) are setup though.

PR #714, opened 2017-09-12
Removing @synchronized(self) block and call from any thread

tonyxiao:
Handles race between

https://github.com/segmentio/analytics-ios/blob/dev/Analytics/Classes/Integrations/SEGIntegrationsManager.m#L499 and https://github.com/segmentio/analytics-ios/blob/dev/Analytics/Classes/Integrations/SEGIntegrationsManager.m#L506 and https://github.com/segmentio/analytics-ios/blob/dev/Analytics/Classes/Integrations/SEGIntegrationsManager.m#L430.

This guarantees that `forwardSelector:arguments:options:` is only ever called from the `_serialQueue`, therefore eliminating the need to have `@synchronized(self)` block. 

Only concern is whether certain integrations require the application state notifications hooks to be called from main thread. If so this can cause an issue. 

cc @f2prateek @ladanazita

### DIFF
--- a/Analytics/Classes/Integrations/SEGIntegrationsManager.m
+++ b/Analytics/Classes/Integrations/SEGIntegrationsManager.m
@@ -425,14 +425,9 @@ static NSString *const kSEGAnonymousIdFilename = @"segment.anonymousId";
 
 - (void)forwardSelector:(SEL)selector arguments:(NSArray *)arguments options:(NSDictionary *)options
 {
-    // If the event has opted in for syncrhonous delivery, this may be called on any thread.
-    // Only allow one to be delivered at a time.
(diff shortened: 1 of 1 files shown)


f2prateek:
> Only concern is whether certain integrations require the application state notifications hooks to be called from main thread. If so this can cause an issue.

I think most do require this.

Do we have another option to deliver these to the main thread, perhaps with `dispatch_get_main_queue`?

tonyxiao:
yes, then the issue is we won't be able to deliver to main thread without causing a potential deadlock (with the `@synchronized` block anyways). On a second look, maybe it's ok for this part of the code to execute concurrently because it's not actually affected by race condition?

PR #715, opened 2017-09-13
Always deliver events asynchronously to integrations.

f2prateek:
Removes the ability for certain events to be delivered synchronously to integrations. 

It would be good to do a audit of integrations which need this functionality and make the fix I suggested for those integrations.

Closes #714.

### DIFF
--- a/Analytics/Classes/Integrations/SEGIntegrationsManager.m
+++ b/Analytics/Classes/Integrations/SEGIntegrationsManager.m
@@ -372,14 +372,14 @@ static NSString *const kSEGAnonymousIdFilename = @"segment.anonymousId";
         if (self.settingsRequest) {
             return;
         }
-        
+
         self.settingsRequest = [self.httpClient settingsForWriteKey:self.configuration.writeKey completionHandler:^(BOOL success, NSDictionary *settings) {
             seg_dispatch_specific_async(_serialQueue, ^{
                 if (success) {
                     [self setCachedSettings:settings];
                 } else {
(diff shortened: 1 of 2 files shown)


tonyxiao:
Btw beware of whitespace changes, they can make PR harder to review.

tonyxiao:
@f2prateek i thought in https://github.com/segmentio/analytics-ios/pull/714 you mentioned that "certain integrations require the application state notifications hooks to be called from main thread"? In this PR we removed the ability to do synchronous delivery IN ADDITION to removing the application state notification calls from main thread. Can we confirm this won't cause issue with bundled integrations?

f2prateek:
Yes that's right. This will cause an issue with integrations, I know for sure we'll need to update the Appboy integration. That's why I suggested the fix in the comments. I'm definitely open to other fixes, and it would be ideal to have a fix that allows for both (but #714 doesn't fix the issue).

tonyxiao:
What if we simply got rid of queue / lock based synchronization for app state notifications?

tonyxiao:
there doesn't seem to be any race conditions which would cause a problem in this particular case. So we can just call integration methods synchronously from the main thread where notifications originate. Without `@sychronized(self)` block things won't block.

f2prateek:
> So we can just call integration methods synchronously from the main thread

Since sometimes integrations haven't been initialized yet, we can't always do that, and still will have to deliver them async. Hence we had the `@ sychronized(self)` block so that when that happens, the main thread and the background thread don't both deliver the event at the same time.

PR #716, opened 2017-09-15
Disable code coverage to fix app submission with Xcode 9

NachoSoto:
**What does this PR do?**
With this change, apps submitted to `TestFlight` using `Xcode 9.0` will fail with the following error:
> Invalid Bundle - Disallowed LLVM instrumentation. Do not submit apps with LLVM profiling instrumentation or coverage collection enabled. Turn off LLVM profiling or code coverage, rebuild your app and resubmit the app.

Nothing to test, as this only affects running tests.

See https://github.com/Carthage/Carthage/issues/2056.

**Questions:**
- Does the docs need an update? No
- Do we need to update engineering / success? I don't know?

@segmentio/gateway

### DIFF
--- a/Analytics.xcodeproj/xcshareddata/xcschemes/Analytics.xcscheme
+++ b/Analytics.xcodeproj/xcshareddata/xcschemes/Analytics.xcscheme
@@ -26,8 +26,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES"
-      codeCoverageEnabled = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES">
       <Testables>
          <TestableReference
             skipped = "NO">


NachoSoto:
Oh, you guys actually use test coverage? Unfortunately, building this project with Carthage would mean it's impossible to submit apps, as explained in that Carthage issue :(

f2prateek:
Yeah we do use it :)

To clarify the impact here, there are 3 days customers can use analytics-ios via Carthage, CocoaPods or integrating the framework manually — and this bug only affects the Carthage installation?

f2prateek:
Also, does this only affect TestFlight submissions, or iTunes Connect uploads as well?

NachoSoto:
> and this bug only affects the Carthage installation?

That's correct.

> Also, does this only affect TestFlight submissions, or iTunes Connect uploads as well?

It's the same thing :) so both.

f2prateek:
Got it, thanks! I'd hate to turn off code coverage entirely . I wonder if we can have it off by default in the source, but do some CI hackery to set `codeCoverageEnabled = "YES"`. Or does that sound like a terrible idea?

f2prateek:
Basically wondering we can run the opposite of this on CI https://github.com/Carthage/Carthage/issues/2056#issuecomment-329430394. Will play around with this approach a bit.

NachoSoto:
Actually I think what we can do is have a separate scheme just for testing, that code coverage can use. I can do that on Monday.

f2prateek:
hey @NachoSoto, would love to to get the alternative fix in! Did you send a similar PR to any other projects with a seperate schema for testing - that way we can just copy that approach here.

f2prateek:
I'm going to close this PR for now, I think disabling code coverage here can have other negative consequences.

The best workaround at the moment  seems to be https://github.com/Carthage/Carthage/issues/2056#issuecomment-329430394. It's not ideal that this has to be done by consumers of the library rather than us handle it.

PR #723, opened 2017-10-14
Format NSDate objects to RFC 3339 compliant strings.

djfink-carglass:
**What does this PR do?**

Fixes and issue where dates were not being formatted correctly before being sent to Segment. The correct way to handle RFC 3339 with NSDateFormatter is detailed here: https://developer.apple.com/library/content/qa/qa1480/_index.html. The one small change is to include milliseconds as they were there previously. This will still be RFC 3339 compliant as can be seen here: https://tools.ietf.org/html/rfc3339#section-5.8. 

N/A

Sends some date data through the SDK and verify it is being formatted correctly.

After talking to Segment support I learned our date traits were not being sent to Redshift because they were not RFC 3339 compliant. The link to the support ticket is here: 
 https://segment.zendesk.com/hc/requests/87566

N/A

N/A

**Questions:**
- Does the docs need an update? No
- Do we need to update engineering / success? No

@segmentio/gateway

### DIFF
--- a/Analytics/Classes/Internal/SEGAnalyticsUtils.m
+++ b/Analytics/Classes/Internal/SEGAnalyticsUtils.m
@@ -19,7 +19,8 @@ NSString *iso8601FormattedString(NSDate *date)
     dispatch_once(&onceToken, ^{
         dateFormatter = [[NSDateFormatter alloc] init];
         dateFormatter.locale = [NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"];
-        dateFormatter.dateFormat = @"yyyy-MM-dd'T'HH:mm:ss.SSSZ";
+        dateFormatter.dateFormat = @"yyyy'-'MM'-'dd'T'HH':'mm':'ss.SSS'Z'";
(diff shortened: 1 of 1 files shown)


f2prateek:
Thanks! Could we add a small test case for `iso8601FormattedString` as well?

f2prateek:
Also wondering, maybe we should just use https://developer.apple.com/documentation/foundation/nsiso8601dateformatter/1643114-init?language=objc directly?

djfink-carglass:
I'd like to keep in the milliseconds as it looks like the segment backend is using this format for date fields it generates like "receivedAt". I didn't see anyway to do this with NSISO8601DateFormatter. NSISO8601DateFormatWithFractionalSeconds looks promising but it is only available in iOS 11.0+

f2prateek:
Ok that's fair, it'd be great to get some simple test cases added here and we can get this merged!

f2prateek:
Awesome, thanks! Just waiting on CI and will merge right after.

djfink-carglass:
Sounds good, thanks!

Issue #743, opened 2018-01-12
sharedAnalytics returns null


louoso:
`+ (instancetype)sharedAnalytics;` returns null if invoked before setup. Obviously, this is not intended to occur, but it would be nice to have the Swift type system enforce this.

f2prateek:
Closed by #744

PR #744, opened 2018-01-12
sharedAnalytics returns null before setup

louoso:
**What does this PR do?**

changes `+ (instancetype)sharedAnalytics;` to `+ (instancetype _Nullable)sharedAnalytics;`

at `+ (instancetype)sharedAnalytics;`

It shouldn't change any functionality so whatever sanity tests are normally run should suffice.

I discovered an instance where an event was attempting to be logged before the client was setup that failed silently. I think this would help prevent that from happening to others.

https://github.com/segmentio/analytics-ios/issues/743

**Questions:**

No

No

¯\_(ツ)_/¯

@segmentio/gateway

### DIFF
--- a/Analytics/Classes/SEGAnalytics.h
+++ b/Analytics/Classes/SEGAnalytics.h
@@ -45,7 +45,7 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @see -setupWithConfiguration:
  */
-+ (instancetype)sharedAnalytics;
++ (instancetype _Nullable)sharedAnalytics;
 
 /*!
  @method


f2prateek:
Thanks!

f2prateek:
So I guess one gripe with this is that it might get annoying to use `SEGAnalytics sharedAnalytics` as now it will always prompt you for a check, even when you know you've already initialized this (as we recommend doing this in didFinishLaunching)

louoso:
In our case we did setup Segment in `didFinishLaunching`, but at some point other code that used segment was inadvertently moved above that setup call. It wasn't obvious that was the case, so we didn't notice until that specific event stopped being logged.

So the tradeoff is how often that happens versus having to change `sharedAnalytics` to `sharedAnalytics?` to maintain current behavior.

gustavosaume:
isn't that why there's an Assert there? I feel there's a clear statement by the library to be initialized before getting the instance. if somehow it's getting called before the developer should get the crash and reorder the calls. 

Not only it's annoying to have to unwrap the library every time you want to track something but if a call happens to be executed before it's gonna get ignored making it harder to figure out why.

Issue #703, opened 2017-07-12
Duplicate definition of category GZIP on interface NSData


n8tr:
If you include both GZIP and segment Analytics pods, you get a duplicate category definition of ```NSData (GZIP)```.  The GZIP library should reserve the right to use the NSData (GZIP) category. Segment should rename their category to SEG_GZIP e.g.
```@interface NSData (SEG_GZIP)```

f2prateek:
Seems reasonable. Happy to accept a PR!

f2prateek:
Tracking internally in JIRA as well — https://segment.atlassian.net/browse/LIB-127

f2prateek:
Closed by #749

PR #748, opened 2018-01-26
Add GZIP podfile dependency

gustavosaume:
**What does this PR do?**
Imports the GZIP library (https://github.com/nicklockwood/GZIP) as a pod, which previously was just added into the Vendor folder

Our project imports the GZIP library as a pod and it creates a version conflict.

#703 

**Questions:**
nope

no unless the originally copied GZIP was customized in some way

no unless the originally copied GZIP was customized in some way

@segmentio/gateway


f2prateek:
Hey @gustavosaume, thanks for the PR! Unfortunately we've found that adding a transitive dependency makes it harder for other customers to adopt the libraries (particular those not using CocoaPods/Carthage). While I'd love to enforce a dependency manager, it's not really feasible for all our customers.

For now, I think the best solution to #703 would be to rename our extension so that it doesn't conflict with the one in the Gzip package.

https://github.com/segmentio/analytics-ios/issues/703#issue-242521507
`@interface NSData (SEG_GZIP)`

gustavosaume:
I'll make a new PR, but I feel there has to be a way, I see many other popular libraries that do this 🤷‍♂️ .

PR #749, opened 2018-01-26
Renames GZIP category to prevent collisions

gustavosaume:
**What does this PR do?**
Renames GZIP category to prevent collisions

A naming collision happens when the Analytics library is used in a project that also imports the GZIP library as a pod.

#703

**Questions:**
- Does the docs need an update? **nope**
- Are there any security concerns? **nope**
- Do we need to update engineering / success? **nope**

@segmentio/gateway


f2prateek:
Thank you @gustavosaume !

PR #751, opened 2018-02-07
Ensure queue is always < 1000 items.

f2prateek:
Previously we didn't cap the queue, and when we added the limit, we only removed the oldest element, rather than fully shrink the queue.

Hence there are cases where the queue may already be larger than 1000 events, and removing a single element at a time isn't as helpful. Now we delete as many events as required to trim the queue size.

Ref: LIB-210

### DIFF
--- a/Analytics/Classes/Integrations/SEGIntegrationsManager.m
+++ b/Analytics/Classes/Integrations/SEGIntegrationsManager.m
@@ -417,7 +417,7 @@ static NSString *const kSEGAnonymousIdFilename = @"segment.anonymousId";
     if ([key isEqualToString:@"Segment.io"]) {
         return YES;
     }
-    
+
     if (plan[@"track"][event]) {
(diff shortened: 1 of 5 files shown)


ladanazita:
We should [update docs](https://segment.com/docs/sources/mobile/ios/#how-does-the-sdk-queue-api-calls-) as well since we are dropping events.

Why not have the logic try to flush the queue when it reaches 1000 and fallback to dropping events?

f2prateek:
The queue already flushes when events > flushCount. If we ever get to this state, it's typically because the device has been offline for an extended period and the last 1000-flushCount attempts have failed (by default flush count is 20, that means the last 9980 attempts have failed). I think it's ok to go ahead to removing the events rather than try to flush the queue yet again. Android also does the same.

I'll add a note in the docs - we were already removing these messages before but looks like we didn't document the old behaviour.

f2prateek:
Docs updating here https://github.com/segmentio/site-docs/pull/2719

PR #765, opened 2018-07-06
Add a `maxQueueSize` configuration option

fathyb:
Ref: LIB-442

### DIFF
--- a/Analytics/Classes/Internal/SEGAnalyticsUtils.h
+++ b/Analytics/Classes/Internal/SEGAnalyticsUtils.h
@@ -8,7 +8,7 @@ NSString *GenerateUUIDString(void);
 // Date Utils
 NSString *iso8601FormattedString(NSDate *date);
 
-void trimQueue(NSMutableArray *array, int size);
+void trimQueue(NSMutableArray *array, NSUInteger size);
 
 // Async Utils
 dispatch_queue_t seg_dispatch_queue_create_specific(const char *label,
--- a/Analytics/Classes/Internal/SEGSegmentIntegration.m
+++ b/Analytics/Classes/Internal/SEGSegmentIntegration.m
@@ -465,9 +465,8 @@ static CTTelephonyNetworkInfo *_telephonyNetworkInfo;
 - (void)queuePayload:(NSDictionary *)payload
 {
     @try {
-        // We only queue upto 1000 items, so trim the queue to 1000-1=999
-        // before we add a new element.
-        trimQueue(self.queue, 999);
+        // Trim the queue to maxQueueSize - 1 before we add a new element.
+        trimQueue(self.queue, self.analytics.configuration.maxQueueSize - 1);
         [self.queue addObject:payload];
         [self persistQueue];
         [self flushQueueByLength];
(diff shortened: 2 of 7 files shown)


Analytics/Classes/Internal/SEGAnalyticsUtils.h
```diff
@@ -8,7 +8,7 @@ NSString *GenerateUUIDString(void);
 // Date Utils
 NSString *iso8601FormattedString(NSDate *date);
 
-void trimQueue(NSMutableArray *array, int size);
+void trimQueue(NSMutableArray *array, NSUInteger size);
```
fathyb: Did this to avoid an useless cast from `NSUInteger` (`long` on 64bits) to `int`

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -461,9 +461,8 @@ - (void)enqueueAction:(NSString *)action dictionary:(NSMutableDictionary *)paylo
 - (void)queuePayload:(NSDictionary *)payload
 {
     @try {
-        // We only queue upto 1000 items, so trim the queue to 1000-1=999
-        // before we add a new element.
-        trimQueue(self.queue, 999);
+        // Trim the queue to maxQueueSize - 1 before we add a new element.
+        trimQueue(self.queue, self.analytics.configuration.maxQueueSize - 1);
```
f2prateek: what if somebody sets `maxQueueSize` to 0?

fathyb: Oh, good catch! I guess we have to choose from one of these behaviours (ordered with what I think is best first) : 
1. forbid using `maxQueueSize = 0` and throw an error in `maxQueueSize` setter if 0 is set
2. when `maxQueueSize == 0` ignore all events
3. when `maxQueueSize == 0` behave like `maxQueueSize == 1`

f2prateek: Option 4: `maxQueueSize == 0` should behave like a no-op, i.e. use the default setting.

Analytics/Classes/SEGAnalyticsConfiguration.h
```diff
@@ -61,6 +61,11 @@ typedef NSMutableURLRequest *_Nonnull (^SEGRequestFactory)(NSURL *_Nonnull);
  */
 @property (nonatomic, assign) NSUInteger flushAt;
 
+/**
+ * The maximum number of items to queue before starting to drop old ones.
+ */
+@property (nonatomic, assign) NSUInteger maxQueueSize;
```
f2prateek: Can we also document the default value here?

Analytics/Classes/Internal/SEGSegmentIntegration.m
f2prateek: Option 5: Document that this must be > 1, and when `maxQueueSize == 0` the behaviour is unspecified. I think this is reasonable for now.

fathyb:
Pushed improved docs @f2prateek

f2prateek:
hey @fathyb - think we might need a more complicated rebase here. I'll re-assign this back to you for now!

PR #767, opened 2018-07-13
Add a `flushInterval` configuration option

fathyb:
Ref: LIB-443

### DIFF
--- a/Analytics/Classes/Internal/SEGSegmentIntegration.m
+++ b/Analytics/Classes/Internal/SEGSegmentIntegration.m
@@ -119,7 +119,11 @@ static BOOL GetAdTrackingEnabled()
 
 - (void)setupFlushTimer
 {
-    self.flushTimer = [NSTimer scheduledTimerWithTimeInterval:30.0 target:self selector:@selector(flush) userInfo:nil repeats:YES];
+    self.flushTimer = [NSTimer scheduledTimerWithTimeInterval:self.configuration.flushInterval
+                                                       target:self
+                                                     selector:@selector(flush)
+                                                     userInfo:nil
+                                                      repeats:YES];
 }
 
 /*
--- a/Analytics/Classes/SEGAnalyticsConfiguration.h
+++ b/Analytics/Classes/SEGAnalyticsConfiguration.h
@@ -62,7 +62,10 @@ typedef NSMutableURLRequest *_Nonnull (^SEGRequestFactory)(NSURL *_Nonnull);
 @property (nonatomic, assign) NSUInteger flushAt;
 
 /**
- * The amount of time to wait before each tick of the flush timer. Smaller values will make events delivered in a more real-time manner and also use more battery. `30` by default.
+ * The amount of time to wait before each tick of the flush timer.
+ * Smaller values will make events delivered in a more real-time manner and also use more battery.
+ * A value smaller than 10 seconds will seriously degrade overall performance.
+ * 30 seconds by default.
  */
 @property (nonatomic, assign) NSTimeInterval flushInterval;
 
--- a/Analytics/Classes/SEGAnalyticsConfiguration.h
+++ b/Analytics/Classes/SEGAnalyticsConfiguration.h
@@ -61,6 +61,10 @@ typedef NSMutableURLRequest *_Nonnull (^SEGRequestFactory)(NSURL *_Nonnull);
  */
 @property (nonatomic, assign) NSUInteger flushAt;
 
+/**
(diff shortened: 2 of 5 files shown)


Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -109,17 +109,44 @@ - (id)initWithAnalytics:(SEGAnalytics *)analytics httpClient:(SEGHTTPClient *)ht
         if ([NSThread isMainThread]) {
             [self setupFlushTimer];
         } else {
-            dispatch_sync(dispatch_get_main_queue(), ^{
+            dispatch_async(dispatch_get_main_queue(), ^{
```
fathyb: This one is interesting. After updating Xcode I started getting random deadlocks when testing, even on a clean repo on `origin/master` without any changes and `DerivedData` nuked.

The debugger showed that two threads (`main` and our serial queue) are locking the main queue : 
screen shot 2018-07-13 at 14 13 34
screen shot 2018-07-13 at 14 13 47

Because we just need to set the timer to let ARC keep a reference on it I replaced the code to use `async`  instead of `sync`, which should prevent any deadlock on this section. What do you think @f2prateek?

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -109,17 +109,44 @@ - (id)initWithAnalytics:(SEGAnalytics *)analytics httpClient:(SEGHTTPClient *)ht
...
                 [self setupFlushTimer];
             });
         }
+        
+        // Observe changes to flushInterval and update the timer
+        [self.configuration addObserver:self
```
fathyb: Figured we might want to observe this, let me know if you think this is overkill.

Analytics/Classes/Internal/SEGSegmentIntegration.m
f2prateek: I actually think we've gone back and forth on this. I can try digging up the history, you might be able to find it via git blame as well.

f2prateek:
This change is a bit bigger than I expected. I think we should narrow it down a bit more, and avoid changing too much logic without scoping out the impact in a bit more detail.

I was thinking all we need to change here would be not hardcode `30` here https://github.com/segmentio/analytics-ios/blob/master/Analytics/Classes/Internal/SEGSegmentIntegration.m#L122, and pull it from the configuration instead.

We can still do bigger changes if we need to but I think it should be a different PR/discussion. What do you think?

fathyb:
@f2prateek you're right, I was a bit worried about that. I pushed a simpler version with no observers and no logic changes

Analytics/Classes/SEGAnalyticsConfiguration.h
```diff
@@ -61,6 +61,10 @@ typedef NSMutableURLRequest *_Nonnull (^SEGRequestFactory)(NSURL *_Nonnull);
  */
 @property (nonatomic, assign) NSUInteger flushAt;
 
+/**
+ * The amount of time to wait before each tick of the flush timer. Smaller values will make events delivered in a more real-time manner and also use more battery. `30` by default.
+ */
+@property (nonatomic, assign) NSTimeInterval flushInterval;
```
f2prateek: Let's document the default value.

f2prateek: Can someone set this to less than  < 0?

Analytics/Classes/SEGAnalyticsConfiguration.h
```diff
@@ -61,6 +61,10 @@ typedef NSMutableURLRequest *_Nonnull (^SEGRequestFactory)(NSURL *_Nonnull);
  */
 @property (nonatomic, assign) NSUInteger flushAt;
 
+/**
+ * The amount of time to wait before each tick of the flush timer. Smaller values will make events delivered in a more real-time manner and also use more battery. `30` by default.
```
f2prateek: 30 seconds will be clearer than just 30 I think

Analytics/Classes/SEGAnalyticsConfiguration.h
f2prateek: I actually think something too low (1ms) may also be problematic. We should just say something like "Don't set this below 10seconds".

fathyb: [According to Apple](https://developer.apple.com/documentation/foundation/nstimer/2091889-scheduledtimerwithtimeinterval?language=objc) :
> **interval**
> > The number of seconds between firings of the timer. If interval is less than or equal to 0.0, this method chooses the nonnegative value of 0.1 milliseconds instead.

I added this note : 
> A value smaller than 10 seconds will seriously degrade overall performance.

PR #768, opened 2018-07-16
Pass E2E credentials using environment variables

fathyb:
Ref: LIB-460

### DIFF
--- a/AnalyticsTests/EndToEndTests.swift
+++ b/AnalyticsTests/EndToEndTests.swift
@@ -4,17 +4,21 @@ import Analytics
 import Alamofire
 import Alamofire_Synchronous
 
-// End to End tests require some private credentials, so we can't embed them in source.
-// On CI, we inject these values with sed (see circle config for exact command).
-let RUN_E2E_TESTS = false
-let WEBHOOK_AUTH_USERNAME = "{WEBHOOK_AUTH_USERNAME}"
-
 func hasMatchingId(messageId: String) -> Bool {
-  if !RUN_E2E_TESTS {
+  // End to End tests require some private credentials, so we can't embed them in source.
+  // On CI, we inject these values with build settings (see Makefile test task).
+  let runE2E = ProcessInfo.processInfo.environment["RUN_E2E_TESTS"]
+
+  if runE2E != "true" {
     return true
   }
 
-  let base64Token = SEGHTTPClient.authorizationHeader(WEBHOOK_AUTH_USERNAME)
+  guard let auth = ProcessInfo.processInfo.environment["WEBHOOK_AUTH_USERNAME"] else {
+    fail("Cannot find webhook username")
+    return false
+  }
+
+  let base64Token = SEGHTTPClient.authorizationHeader(auth)
 
   let headers: HTTPHeaders = [
     "Authorization": "Basic \(base64Token)",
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,9 +1,10 @@
 SDK ?= "iphonesimulator"
 DESTINATION ?= "platform=iOS Simulator,name=iPhone X"
 PROJECT := Analytics
-XC_ARGS := -scheme $(PROJECT) -workspace $(PROJECT).xcworkspace -sdk $(SDK) -destination $(DESTINATION) GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=YES
-XC_BUILD_ARGS := ONLY_ACTIVE_ARCH=NO
-XC_TEST_ARGS := GCC_GENERATE_TEST_COVERAGE_FILES=YES
+TEST_PROJECT := $(PROJECT)Tests
+XC_ARGS := -workspace $(PROJECT).xcworkspace -sdk $(SDK) -destination $(DESTINATION) GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=YES
+XC_BUILD_ARGS := -scheme $(PROJECT) ONLY_ACTIVE_ARCH=NO
+XC_TEST_ARGS := -scheme $(TEST_PROJECT) GCC_GENERATE_TEST_COVERAGE_FILES=YES RUN_E2E_TESTS=$(RUN_E2E_TESTS) WEBHOOK_AUTH_USERNAME=$(WEBHOOK_AUTH_USERNAME)
```
f2prateek: what happens if the environment (e.g. a local machine) does not have this (`RUN_E2E_TESTS`) set? does it just skip the tests and still work?

f2prateek: I tested locally and it seems to work as expected!

AnalyticsTests/EndToEndTests.swift
```diff
@@ -4,17 +4,21 @@ import Analytics
...
+  // End to End tests require some private credentials, so we can't embed them in source.
+  // On CI, we inject these values with sed (see circle config for exact command).
+  let runE2E = ProcessInfo.processInfo.environment["RUN_E2E_TESTS"]
+
+  if runE2E != "true" {
     return true
   }
 
-  let base64Token = SEGHTTPClient.authorizationHeader(WEBHOOK_AUTH_USERNAME)
+  guard let auth = ProcessInfo.processInfo.environment["WEBHOOK_AUTH_USERNAME"] else {
```
f2prateek: nice!

AnalyticsTests/EndToEndTests.swift
```diff
@@ -4,17 +4,21 @@ import Analytics
...
 
-// End to End tests require some private credentials, so we can't embed them in source.
-// On CI, we inject these values with sed (see circle config for exact command).
-let RUN_E2E_TESTS = false
-let WEBHOOK_AUTH_USERNAME = "{WEBHOOK_AUTH_USERNAME}"
-
 func hasMatchingId(messageId: String) -> Bool {
-  if !RUN_E2E_TESTS {
+  // End to End tests require some private credentials, so we can't embed them in source.
+  // On CI, we inject these values with sed (see circle config for exact command).
```
f2prateek: Lets update the docs here now that we use build settings instead?

Makefile
fathyb: Right, it should only run it when the `RUN_E2E_TESTS` environment variable is `true`

PR #771, opened 2018-07-25
Custom name/properties for auto screen reporting

### LINKED ISSUE #654: Screen tracking doesn't always produce desired name for UITabBarController and UISplitViewController
It seems that the output of `seg_topViewController` doesn't always pick the right view top view controller especially for instances when we have nested child view controllers.

Is this particular case we have our navigation controller nested within a `RootViewController` which results in all our screen tracking events being called: "Root"

Happy to either PR to produce better results, or perhaps inclined to not use provided `recordScreenViews` view controller swizzler. 

Thanks.

humblehacker:
**What does this PR do?**

Adds the ability to specify a custom name and custom properties during
automatic screen reporting. 

Also fixes `seg_topViewController()` for `UITabBarController` and custom
container view controllers.

`UIViewController.seg_viewDidAppear()`

* Create a UIViewController, and conform to the new `SEGScreenReporting`
protocol.  
* Implement `seg_trackScreen` to override the default screen tracking.  Your implementation should call `seg_screen` with custom values for the `screen`, `properties` and/or `options` arguments.
* Implement `seg_mainViewController` to specify which child view
controller of a custom container view controller should be referenced
when determining the top view controller (`seg_topViewController()`)

None

Fixes #654

No UI/UX change

**Questions:**
- Do the docs need an update?

    Yes

    Not that I'm aware of 

    Not sure what this means

@segmentio/gateway

### DIFF
--- a/Analytics.xcodeproj/project.pbxproj
+++ b/Analytics.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		3481C14AEC76E5A8DA64DA59 /* Pods_AnalyticsTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 89033CBF22319674E6CE7E61 /* Pods_AnalyticsTests.framework */; };
+		630FC8EA2107F2A500A759C5 /* SEGScreenReporting.h in Headers */ = {isa = PBXBuildFile; fileRef = 5AF0EDFEDC0EE79A0A0EB713 /* SEGScreenReporting.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		6E265C791FB1178C0030E08E /* IntegrationsManagerTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6E265C781FB1178C0030E08E /* IntegrationsManagerTest.swift */; };
 		6EEC1C712017EA370089C478 /* EndToEndTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6EEC1C702017EA370089C478 /* EndToEndTests.swift */; };
 		EA88A5981DED7608009FB66A /* SEGSerializableValue.h in Headers */ = {isa = PBXBuildFile; fileRef = EA88A5971DED7608009FB66A /* SEGSerializableValue.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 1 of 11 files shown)


humblehacker:
Is there anything in this PR that you'd like to see changed?

f2prateek:
hey @humblehacker sorry for the delay. 

I do have a question before diving too much into the implementation. If a customer has to write code to get this to work (i.e. they have to implement the new protocol), wouldn't it be just as easy to call `[[SEGAnalytics sharedAnalytics] screen:...]`?

humblehacker:
Thanks for the response. Yes, they could just call `screen` themselves, but they'd have to either make that call manually for every view controller (losing the automatic screen tracking), or reimplement the `UIViewController` swizzling already performed by the Segment SDK.  If they try to do both (use automatic tracking as well as manual calls to `screen`), they would get multiple screen events for the same view controller presentation.

f2prateek:
So far automatic screen recording is very binary - you either use it or you don't. I would like to avoid complicating the feature set too much.

But agreed it would be nice to have a way to customize the calls somehow.

Maybe.... instead of a protocol with three properties, we could have the protocol be something like this:

```
@protocol SEGScreenReporting
- (void)seg_screen;
@end
```

Then if our library detects this, instead of calling screen directly, it calls the `seg_screen` method and the customer can make the actual screen call.

Another similar idea would be:

````
@protocol SEGScreenReporting
@optional @property (readonly, nullable) NSString * seg_screenOptOut;
@end
```

Then if our library detects this, it doesn't call screen at all and lets the customer record screen manually.

It's a bit hard to decide since we don't have a lot of customers who have really asked for this yet. 

I would suggest we leave this open so folks can leave some feedback on this before we decide to proceed. How does that sound?

humblehacker:
That's a reasonable compromise -- accomplishes the same goal with slightly less complexity.  I prefer your first suggestion over the second.  Though maybe `seg_trackScreen` might be a better name?

This eliminates two properties from the protocol, but the third property (`seg_mainViewController`) serves a different purpose of allowing custom container view controllers to participate in automatic screen tracking (related to #654).  Thoughts?

I'm fine with leaving this open for a while to gather feedback.

humblehacker:
Hello again.  Is there anything I can do to help move this along?

f2prateek:
Sorry for the delay! I think this looks reasonable to me - could we add some tests?

I don't anything else is blocking this - cc @fathyb to hear your thoughts too!

timbroder:
Is there anything left to move this along? We have a tab bar app and would like to use auto tracking :)

itsallmememe:
Is there any way to do auto-screen tracking with container views? I can't seem to find any information about this feature, and this PR sounds like it will fix the issue.

humblehacker:
I've rebased against master and added some tests.  Let me know if there's anything else I can do.

humblehacker:
Not sure why, but `AnalyticsTests.swift` hangs on my machine, and it looks like the same on CircleCI.  Nothing to do with this PR though.  I validated the same issue occurs against `segmentio:master`.

humblehacker:
I realize it's been a while since this PR was opened. Is there any chance for its acceptance?

bsneed:
Thanks @humblehacker !!  Sorry for the extreme delay getting this in.  I did the fixups needed and made sure the tests were kosher.  Thanks again!!

bsneed:
ps. #884 is the PR for this in progress release where your changes will come in.

kowongh:
Is there a way to not have the `title` of the viewController not override the class name?

bsneed:
@kowongh there's not currently a way to do that.  You wanna throw up a PR with the logic changes you'd like and I can spin another release once we've agreed?

kowongh:
@bsneed Thanks Brandon. Here's my requested fix for making the naming convention more consistent for the recordScreenViews feature. https://github.com/segmentio/analytics-ios/pull/885

simonbromberg:
> Your implementation should call seg_screen with custom values for the screen, properties and/or options arguments

I don't see `seg_screen` has this changed?

PR #772, opened 2018-07-30
Cache 4xx HTTP responses

fathyb:
Ref: LIB-468

### DIFF
--- a/Analytics/Classes/Internal/SEGHTTPClient.m
+++ b/Analytics/Classes/Internal/SEGHTTPClient.m
@@ -160,10 +160,14 @@
 
         NSInteger code = ((NSHTTPURLResponse *)response).statusCode;
         if (code > 300) {
+            // Look for a non-cached valid 4xx response
             if(response && data && code < 500 && ![self.cache cachedResponseForRequest:request]) {
-                // Cache 4xx responses
                 NSCachedURLResponse* cached = [NSCachedURLResponse.alloc initWithResponse:response
                                                                                      data:data];
+                
+                // Manually store the response onto the cache.
+                // self.genericSession will take care of storing cacheable 2xx responses,
+                // and checking the cache next time dataWithRequest is called.
                 [self.cache storeCachedResponse:cached forRequest:request];
             }
 
--- a/Analytics/Classes/Internal/SEGHTTPClient.m
+++ b/Analytics/Classes/Internal/SEGHTTPClient.m
@@ -28,8 +28,15 @@
         } else {
             self.requestFactory = requestFactory;
         }
-        _sessionsByWriteKey = [NSMutableDictionary dictionary];
+        
+        NSUInteger megabyte = 1024 * 1024;
         NSURLSessionConfiguration *config = [NSURLSessionConfiguration defaultSessionConfiguration];
+
+        _sessionsByWriteKey = [NSMutableDictionary dictionary];
+        _cache = [NSURLCache.alloc initWithMemoryCapacity:1 * megabyte
+                                             diskCapacity:5 * megabyte
+                                                 diskPath:@"seg_http_cache"];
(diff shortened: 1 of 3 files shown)


f2prateek:
Can you describe the effect of this change a bit more? It's not clear why this we are adding code to cache 4xx's (and why not other response codes).

Analytics/Classes/Internal/SEGHTTPClient.m
```diff
@@ -153,6 +160,13 @@ - (NSURLSessionDataTask *)settingsForWriteKey:(NSString *)writeKey completionHan
 
         NSInteger code = ((NSHTTPURLResponse *)response).statusCode;
         if (code > 300) {
+            if(response && data && code < 500 && ![self.cache cachedResponseForRequest:request]) {
+                // Cache 4xx responses
+                NSCachedURLResponse* cached = [NSCachedURLResponse.alloc initWithResponse:response
+                                                                                     data:data];
+                [self.cache storeCachedResponse:cached forRequest:request];
```
f2prateek: how long is ths cached for?

f2prateek:
I think there might be a slight side effect here. Currently we should be caching 2xx responses as indicated by the cache headers.

```
$ curl -I http://cdn.segment.com/v1/projects/MbC3FWS4XIWsaUhPusDBdC0kuLkhFwax/settings
...
Cache-Control: public, max-age=3600
...
```

I think with this change we would only be caching 4xx responses, and not caching the 2xx responses at all.

fathyb:
`NSURLSession` should already handle HTTP cache headers. This is configured by [`NSURLSessionConfiguration.requestCachePolicy`](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1411655-requestcachepolicy) which is [`NSURLRequestUseProtocolCachePolicy `](https://developer.apple.com/documentation/foundation/nsurlrequestcachepolicy/nsurlrequestuseprotocolcachepolicy?language=objc) by default. We were already caching 2xx responses per iOS default behavior, this only adds a manual caching path for 4xx responses.

A slight change is that we now manually configure the cache size instead of using the shared one : 
https://github.com/segmentio/analytics-ios/blob/1d3c9e514876d6e6a82ede2c932367eb845996b2/Analytics/Classes/Internal/SEGHTTPClient.m#L36-L38

I've just noticed that it's smaller than the default one (disk: 20M, mem: 40M). I've changed it so they have the same size.

Analytics/Classes/Internal/SEGHTTPClient.m
fathyb: It'll only be one hour (according to `max-age`) and after it'll start issuing a `HEAD` request to check `ETag`. We can also store it indefinitely if needed

f2prateek:
I see - and does the cache cache 4xx requests? `curl -I http://cdn.segment.com/v1/projects/foo/settings` does return max-age

fathyb:
Sorry, I've buried your review comment by commiting. I've answered :
> It'll only be one hour (according to `max-age`) and after it'll start issuing a `HEAD` request to check `ETag`. We can also store it indefinitely if needed

There is a cool decision tree in [`NSURLRequestUseProtocolCachePolicy`](https://developer.apple.com/documentation/foundation/nsurlrequestcachepolicy/nsurlrequestuseprotocolcachepolicy?language=objc) : 

http-cache


We can override `max-age` before saving the request to something like 1 year. What does seem reasonable to you?

f2prateek:
I think we should respect the cache headers (so the server can control this if needed).

Sorry my original question was - does the *default* cache cache 4xx responses?

fathyb:
Yes, it does.

The built-in network inspector reported cached requests, after checking using a TCP proxy I found that 4xx responses are cached. I'm closing this as this is not needed, sorry!

PR #773, opened 2018-08-21
Declare framework dependencies in podspec

jgavris:
Users may want to build the pod with Clang modules disabled -fno-modules because they are using ccache which doesn't support -fmodules. When modules are disabled, auto-linking of framework modules is also disabled, so these dependencies need to be expressed explicitly.

**What does this PR do?**

Declares all the frameworks the pod depends on explicitly in the podspec, so that users may disable 'Clang Modules' in order to build the Objective-C code using `ccache` for faster builds. When `CLANG_ENABLE_MODULES` is off, automatic framework linking is also implicitly turned off, so without this declaration the pod no longer links.

Review that all frameworks listed are used in the pod.

Perform `bundle exec pod install` of this podspec and ensure the pod still builds in the target.

**Questions:**

No.

No, these are all Apple provided iOS frameworks.

No.

@segmentio/gateway

### DIFF
--- a/Analytics.podspec
+++ b/Analytics.podspec
@@ -17,7 +17,7 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = '7.0'
   s.tvos.deployment_target = '9.0'
 
-  s.framework = 'Security'
+  s.frameworks = 'CoreTelephony', 'Security', 'StoreKit', 'SystemConfiguration', 'UIKit'
 
   s.source_files = [
     'Analytics/Classes/**/*',


f2prateek:
@jgavris thanks! I had a question - how did you figure out which frameworks to add? Was it manual, or can we use a tool to do this?

I'll let @fathyb review some of the details as well, but

jgavris:
I actually just iteratively compiled the project with `CLANG_ENABLE_MODULES=NO` in the pod target and looked at which frameworks provided the missing symbols. 

You could also do something like 

```shell
λ grep -hr "import <" . | sort -n | uniq
grep: ./.clang-format: No such file or directory
#import <AdSupport/ASIdentifierManager.h>
#import <Analytics/Analytics.h>
#import <Analytics/NSData+SEGGZIP.h>
#import <Analytics/SEGAES256Crypto.h>
#import <Analytics/SEGAnalytics.h>
#import <Analytics/SEGAnalyticsUtils.h>
#import <Analytics/SEGFileStorage.h>
#import <Analytics/SEGIntegrationsManager.h>
#import <Analytics/SEGSerializableValue.h>
#import <Analytics/SEGStoreKitTracker.h>
#import <Analytics/SEGUserDefaultsStorage.h>
#import <Analytics/UIViewController+SEGScreen.h>
#import <CommonCrypto/CommonCryptor.h>
#import <CommonCrypto/CommonKeyDerivation.h>
#import <CoreTelephony/CTCarrier.h>
#import <CoreTelephony/CTTelephonyNetworkInfo.h>
#import <Foundation/Foundation.h>
#import <StoreKit/StoreKit.h>
#import <SystemConfiguration/SystemConfiguration.h>
#import <UIKit/UIKit.h>
#import <arpa/inet.h>
#import <dlfcn.h>
#import <ifaddrs.h>
#import <netdb.h>
#import <netinet/in.h>
#import <netinet6/in6.h>
#import <objc/runtime.h>
#import <sys/socket.h>
#import <zlib.h>
```

f2prateek:
Thanks - this makes it much easier to verify if we missed something.

jgavris:
I will add a change to the podfile that should test this on CI.

Issue #518, opened 2016-03-19
Seeing lockups writing to NSUserDefaults


gestrich:
We are seeing NSUserDefaults locking up for a small number of our users after integrating Segment. One thread is waiting on the mutex for CFPrefsSource, but no other threads seem to have acquired the mutex. This is the only thread writing/reading to NSUserDefaults. Then, if another thread tries to write to  NSUserDefaults from the main thread, the app will lock up. Any idea why this call is getting stuck? We can't reproduce it but the users that have the problem have it consistently until we disable segment analytics for them. We assume maybe there is an issue with the data attempting to be written (ex mutable data in the structure?)

Thread 15:
0   libsystem_kernel.dylib               0x00000001835a7f90 __psynch_mutexwait + 8
1   CoreFoundation                       0x0000000183a4260c -[CFPrefsSource setValue:forKey:] + 44
2   CoreFoundation                       0x0000000183a41bbc +[CFPrefsSource withSourceForIdentifier:user:byHost:container:perform:] + 836
3   CoreFoundation                       0x0000000183a453cc _CFPreferencesSetValueWithContainer + 300
4   Foundation                           0x00000001842f1b78 -[NSUserDefaults(NSUserDefaults) setObject:forKey:] + 52
5   ForeFlight                           0x00000001010fcba0 -[SEGSegmentIntegration persistQueue] (SEGSegmentIntegration.m:583)
6   ForeFlight                           0x00000001010fb5c8 -[SEGSegmentIntegration queuePayload:] (SEGSegmentIntegration.m:397)
7   ForeFlight                           0x00000001010fb4a4 __71-[SEGSegmentIntegration enqueueAction:dictionary:context:integrations:]_block_invoke (SEGSegmentIntegration.m:389)
8   libdispatch.dylib                    0x0000000183459630 _dispatch_call_block_and_release + 20
9   libdispatch.dylib                    0x00000001834595f0 _dispatch_client_callout + 12
10  libdispatch.dylib                    0x0000000183465634 _dispatch_queue_drain + 860
11  libdispatch.dylib                    0x000000018345d0f4 _dispatch_queue_invoke + 460
12  libdispatch.dylib                    0x00000001834595f0 _dispatch_client_callout + 12
13  libdispatch.dylib                    0x0000000183467a88 _dispatch_root_queue_drain + 2136
14  libdispatch.dylib                    0x0000000183467224 _dispatch_worker_thread3 + 108
[...]

PR #784, opened 2018-09-19
Fix flaky tests

fathyb:
This PR fixes 3 common source of failure for our test :

1. ~Deadlock : sometimes the test runner freeze and needs a total clean to be run. We fix this by removing the need to call the main thread in SegmentIntegration~ Moved to #784
2. Missing iOS simulator : Adding a command to list the simulators seems to magically fix the issue.
3. Flaky maxQueueSize test : I rewrote this one to be more stable

We ran 21+ successful and consecutive test iterations : https://circleci.com/gh/segmentio/workflows/analytics-ios/tree/fix%2Fflaky-test

Ref : LIB-543

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,7 @@ jobs:
       xcode: "9.4.1"
     steps:
       - checkout
+      - run: xcrun simctl list
       - run:
           name: Install build dependencies
           command: |
--- a/AnalyticsTests/AnalyticsTests.swift
+++ b/AnalyticsTests/AnalyticsTests.swift
@@ -121,19 +121,18 @@ class AnalyticsTests: QuickSpec {
       let integration = analytics.test_integrationsManager()?.test_segmentIntegration()
       expect(integration).notTo(beNil())
       
-      var sent = 0
-
       analytics.flush()
-      integration?.test_dispatchBackground {
-        if let count = integration?.test_queue()?.count {
-          sent = count
-        }
-        else {
-          sent = -1
+      waitUntil(timeout: 60) {done in
+        let queue = DispatchQueue(label: "test")
+        
+        queue.async {
+          while(integration?.test_queue()?.count != max) {
+            sleep(1)
+          }
+
+          done()
         }
       }
-      
-      expect(sent).toEventually(equal(max))
     }
 
     it("protocol conformance should not interfere with UIApplication interface") {
--- a/AnalyticsTests/EndToEndTests.swift
+++ b/AnalyticsTests/EndToEndTests.swift
@@ -65,9 +65,9 @@ class AnalyticsE2ETests: QuickSpec {
 
       analytics.track("E2E Test", properties: ["id": uuid])
 
-      self.waitForExpectations(timeout: 20)
+      self.waitForExpectations(timeout: 3 * 60)
 
-      for _ in 1...5 {
+      for _ in 1...3 * 30 {
         sleep(2)
         if hasMatchingId(messageId: uuid) {
           return


.circleci/config.yml
```diff
@@ -6,6 +6,7 @@ jobs:
       xcode: "9.4.1"
     steps:
       - checkout
+      - run: xcrun simctl list
```
f2prateek: do we still need this?

Analytics/Classes/Internal/SEGSegmentIntegration.m
```diff
@@ -106,26 +106,18 @@ - (id)initWithAnalytics:(SEGAnalytics *)analytics httpClient:(SEGHTTPClient *)ht
...
         }];
 
-        if ([NSThread isMainThread]) {
-            [self setupFlushTimer];
-        } else {
-            dispatch_sync(dispatch_get_main_queue(), ^{
-                [self setupFlushTimer];
-            });
-        }
+        self.flushTimer = [NSTimer timerWithTimeInterval:self.configuration.flushInterval
```
f2prateek: do we need to check which thread we are on?

AnalyticsTests/AnalyticsTests.swift
```diff
@@ -121,19 +121,18 @@ class AnalyticsTests: QuickSpec {
...
-      var sent = 0
-
       analytics.flush()
-      integration?.test_dispatchBackground {
-        if let count = integration?.test_queue()?.count {
-          sent = count
-        }
-        else {
-          sent = -1
+      waitUntil(timeout: 60) {done in
```
nhi-nguyen: is this the timeout to check for the messages in webhook? if so, you may want to increase it to 3 minutes as 1 minute is sometimes not enough to find them in webhook (3 mins is what library-e2e-tester currently does).

.circleci/config.yml
fathyb: Since I added `xcrun simctl list` we've miraculously never had the error again.  If I remove it we get the missing simulator error sporadically. This could also help us debug it next time.

Analytics/Classes/Internal/SEGSegmentIntegration.m
fathyb: No, we manually add the timer in `NSRunLoop.mainRunLoop` so it'll always be scheduled to run on the main thread:

https://github.com/segmentio/analytics-ios/blob/9f0eecc9846ed123f378ca87bca1afad3053b68e/Analytics/Classes/Internal/SEGSegmentIntegration.m#L115-L116

We needed to check before because we used `scheduledTimerWithTimeInterval` instead of `timerWithTimeInterval`

AnalyticsTests/AnalyticsTests.swift
fathyb: It isn't this one, but I pushed a new commit to increase the webhook timeout to 3 minutes - thanks!

Analytics/Classes/Internal/SEGSegmentIntegration.m
f2prateek: We've had some issues around concurrency - particularly as the threads we initialize integrations is not guaranteed.  

 This causes some integrations and customers to rely on implementation details 
 that have caused a few issues in the past (https://github.com/segmentio/analytics-ios/issues/518, https://github.com/segmentio/analytics-ios/pull/714, https://github.com/segmentio/analytics-ios/issues/683) so I'd like to be a bit careful rolling this out. If you're confident in these changes as is  - let's plan to cut this as a beta release to get a bit more testing through this change!

fathyb: Thanks for pointing this out! If customers are relying on this kind of implementation detail I think it’d be better to spend more time looking into the impact. Moving this to #785.

PR #785, opened 2018-09-21
Fix GCD mutual dependency

fathyb:
I'm putting this in WIP until we're confident this will not result in regressions.

---

Sometimes the test runner freeze and needs a total clean to be run. We fix this by removing the need to call the main thread in SegmentIntegration by using the proper `NSRunLoop` API.

[Prateek noted](https://github.com/segmentio/analytics-ios/pull/784#discussion_r219051393) :
> We've had some issues around concurrency - particularly as the threads we initialize integrations is not guaranteed.
>
> This causes some integrations and customers to rely on implementation details
that have caused a few issues in the past (#518, #714, #683) so I'd like to be a bit careful rolling this out.


f2prateek:
I think some of it actually might have been due to bugs on our side.

but i'd be ok shipping this in the next beta.

Issue #753, opened 2018-03-02
Ability to set anonymousId removed in 3.6.0


jonah-williams:
I stumbled across an unexpected change in behavior after upgrading across 3.6.0.

In https://github.com/segmentio/analytics-ios/compare/3.5.7...3.6.0#diff-90ee8062efa95f661137bf2685fd0256L271 we can see that `identify` started constructing a `SEGIdentifyPayload` with a nil `anonymousId` (and that remains true today in https://github.com/segmentio/analytics-ios/blob/master/Analytics/Classes/SEGAnalytics.m#L203) We used to be able to specify an anonymous id in the `options` argument.

https://github.com/segmentio/analytics-ios/blob/master/Analytics/Classes/Integrations/SEGIntegrationsManager.m#L155 appears to retain some of this behavior which checks `SEGIdentifyPayload.options` but that can now only ever contain `context` and `integrations` keys via https://github.com/segmentio/analytics-ios/blob/master/Analytics/Classes/Integrations/SEGIntegrationsManager.m#L526

I had found the option to specify an anonymous id useful because it allowed me to keep identifiers for crash logging and analytics in sync so I could correlate them as needed. In my case I can switch to allowing Segment to be the authority over my anonymous ids but maybe there should be some option to explicitly set this value, either via `identify` or otherwise?

f2prateek:
Ah yep, this looks like a bug as part of the middleware changes. I think the fix should be to move that logic from `SEGIntegrationsManager` upto `SEGAnalytics.

f2prateek:
I don't have an ETA, but I've triaged the issue internally in our JIRA https://segment.atlassian.net/browse/LIB-266

PR #790, opened 2018-09-25
Make setting the anonymousId public

gabek:
I'm sure there's a reason why saving the anonymousId isn't made public, but I'm not sure what it is, so I'm submitting this PR if, for nothing else, to re-spur the conversation.

**What does this PR do?**
Allows setting of the anonymousId in the integrations manager.

```
segment.getAnonymousId() -> "abc123"
segment.setAnonymousId("test123")
segment.getAnonymousId() -> "test123"
```
Validate in the Segment debugger that the correct anonymous Id comes through.

https://github.com/segmentio/analytics-ios/issues/753

**Questions:**
- Does the docs need an update? Yes
- Are there any security concerns? I'm not aware of any,
- Do we need to update engineering / success?  *shrug*


fathyb:
Fixed with #799

PR #792, opened 2018-10-05
Tvos write to file fix and CoreTelephony Fix

krzemienski:
So we have been having some issues with the user defaults running out of storage space based on other libs for tvos using it on top of segment, you can get around the concept of userdefaults by using the cache directory rather than what you had for ios in particular. 

https://forums.developer.apple.com/thread/20934

### DIFF
--- a/Analytics.podspec
+++ b/Analytics.podspec
@@ -17,7 +17,8 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = '7.0'
   s.tvos.deployment_target = '9.0'
 
-  s.frameworks = 'CoreTelephony', 'Security', 'StoreKit', 'SystemConfiguration', 'UIKit'
+  s.ios.frameworks = 'CoreTelephony', 'Security', 'StoreKit', 'SystemConfiguration', 'UIKit'
+  s.tvos.frameworks = 'Security', 'StoreKit', 'SystemConfiguration', 'UIKit'
 
   s.source_files = [
     'Analytics/Classes/**/*',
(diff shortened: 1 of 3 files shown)


krzemienski:
Actually just noted that Caches is not guaranteed. So in our case id prefer to run a small flush size and large queue size, and run the risk of loosing events than crashing on tryign to insert events into libs. Thoughts? Here was our crash reports and problems: http://crashes.to/s/aff4081f5e0

f2prateek:
@fathyb could you take a look at this one?

f2prateek:
@krzemienski one suggestion - could you split out these into two PRs? one for the core telephony fix and one for the crash fix? This way we can merge and review them separately.

krzemienski:
@f2prateek yes no problem! Just a note unit tests do not run for tvos as well. Ill try and fix that as well but we different options there. Either make a new test target for tvos thats a dup of the ios ones. Have you guys never actually ran the tests against tvos? id like to see what the potential loss on cache directory is. I use it for plenty of things in our prod app and have never had it removed from me. 

I think if we really wanted to gurantee things we could in theory just put the queue in the cache and ensure things like Anonymous id is still in UserDefaults since its very little space.

f2prateek:
That's a good call out - currently we actually only rely on manual testing for tvOS. I think it could be worth the effort to get automated tests running on tvOS as well.

iMacHumphries:
:+1: for this PR. Getting the same crash for 3 apps. Nice work @krzemienski this looks perfect!
cc: @f2prateek @fathyb

krzemienski:
@iMacHumphries thanks you were able to confirm that the cache does still exist though? I havnt looked at this in a long time, but we were able to see all our issues disappear I no longer work on the appletv code base but happy to make any changes needed to see this merged @f2prateek @fathyb please let me know!

iMacHumphries:
@krzemienski The cache is working on simulator. I'm double checking on an actual device because permissions can be different. I think this PR could also remove SEGUserDefaultsStorage which appears to not be used anymore after your change :+1:

iMacHumphries:
Actually I wonder if we should migrate the data in SEGUserDefaultsStorage to NSCachesDirectory for Apple TV 🤔 looks like when I first ran this on device I got: "WARNING: No data file for key" and then a new anonymousId was created.

krzemienski:
Makes sense, I’ll have to break out the MacBook and pull down this branch again, ever since I switched to encoding things I run Linux now...I’ve been wanting to say hi to Xcode again anyways!

iMacHumphries:
@krzemienski I created a PR to resolve this here: #861

PR #795, opened 2018-10-15
Remove -beta suffix from CFBundleShortVersionString

### LINKED ISSUE #794: Builds using 3.7.0-beta cannot be submitted to the App Store
Here's the error I just got from uploading a build of ours that uses 3.7.0-beta from Carthage.

```
[17:00:04]: ERROR ITMS-90060: "This bundle is invalid. The value for key CFBundleShortVersionString '3.7.0-beta' in the Info.plist file must be a period-separated list of at most three non-negative integers."
Return status of iTunes Transporter was 1: ERROR ITMS-90060: "This bundle is invalid. The value for key CFBundleShortVersionString '3.7.0-beta' in the Info.plist file must be a period-separated list of at most three non-negative integers."
The call to the iTMSTransporter completed with a non-zero exit status: 1. This indicates a failure.
```

This means that any Swift 4 project that uses this repo with Carthage cannot submit their application to the App Store.

The fix should be to remove the `-beta` suffix from the version in the Info.plist.

eliperkins:
**What does this PR do?**

This allows for the framework to be used in Swift 4 projects who consume this framework from Carthage, to be submitted to the App Store.

Closes #794

By reading #794

Create a project with Swift 4, consume this framework with Carthage, and then submit it to the App Store 😅

#794

#794

### DIFF
--- a/Analytics/Info.plist
+++ b/Analytics/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.7.0-beta</string>
+	<string>3.7.0</string>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>NSPrincipalClass</key>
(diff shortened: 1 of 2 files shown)


eliperkins:
Alternatively, releasing 3.7.0 will fix #794 but that's beyond the scope of what I can submit a PR for 😅

f2prateek:
@fathyb This makes it challenging to see what version of the library is being used since this will report a beta release version the same as a stable release. We use this data to decide when to promote a beta release to a stable release.

PR #798, opened 2018-11-14
Add deep links redaction

fathyb:
Ref: LIB-686

### DIFF
--- a/Analytics/Classes/Internal/SEGUtils.m
+++ b/Analytics/Classes/Internal/SEGUtils.m
@@ -34,4 +34,62 @@
     return plist;
 }
 
+
++(id)traverseJSON:(id)object andReplaceWithFilters:(NSDictionary<NSString*, NSString*>*)patterns
+{
+    if (object == nil || object == NSNull.null || [object isKindOfClass:NSNull.class]) {
+        return object;
+    }
+    
+    if ([object isKindOfClass:NSDictionary.class]) {
+        NSDictionary* dict = object;
+        NSMutableDictionary* newDict = [NSMutableDictionary dictionaryWithCapacity:dict.count];
+        
+        for (NSString* key in dict.allKeys) {
+            newDict[key] = [self traverseJSON:dict[key] andReplaceWithFilters:patterns];
+        }
+        
+        return newDict;
+    }
+    
+    if ([object isKindOfClass:NSArray.class]) {
+        NSArray* array = object;
+        NSMutableArray* newArray = [NSMutableArray arrayWithCapacity:array.count];
+        
+        for (int i = 0; i < array.count; i++) {
+            newArray[i] = [self traverseJSON:array[i] andReplaceWithFilters:patterns];
+        }
+        
+        return newArray;
+    }
+
+    if ([object isKindOfClass:NSString.class]) {
+        NSError* error = nil;
+        NSMutableString* str = [object mutableCopy];
+        
+        for (NSString* pattern in patterns) {
+            NSRegularExpression* re = [NSRegularExpression regularExpressionWithPattern:pattern
+                                                                                options:0
+                                                                                  error:&error];
+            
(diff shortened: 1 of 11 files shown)


Analytics/Classes/Internal/SEGUtils.m
```diff
@@ -34,4 +34,62 @@ + (id _Nullable)plistFromData:(NSData *_Nonnull)data
...
+            
+            NSInteger matches = [re replaceMatchesInString:str
+                                                   options:0
+                                                     range:NSMakeRange(0, str.length)
+                                              withTemplate:patterns[pattern]];
+            
+            if (matches > 0) {
+                SEGLog(@"%@ Redacted value from action: %@", self, str);
+                
+                return str;
```
nhi-nguyen: It looks like we're trying to match multiple patterns, but here we return immediately after the first match? Should this return statement be outside the `for` loop?

AnalyticsTests/AnalyticsTests.swift
```diff
@@ -170,6 +170,29 @@ class AnalyticsTests: QuickSpec {
...
+      analytics.configuration.payloadFilters["(myapp://auth\\?token=)([^&]+)"] = "$1((redacted/my-auth))"
+      analytics.configuration.trackDeepLinks = true
+      analytics.open(URL(string: "myapp://auth?token=hastoberedacted&other=stuff")!, options: [:])
+      
+      
+      let event = testMiddleware.lastContext?.payload as? SEGTrackPayload
+      expect(event?.event) == "Deep Link Opened"
+      expect(event?.properties?["url"] as? String) == "myapp://auth?token=((redacted/my-auth))&other=stuff"
+    }
   }
```
nhi-nguyen: can we have tests to make sure we're traversing json correctly, especially in the case of nested arrays/dicts?

Analytics/Classes/Internal/SEGUtils.m
fathyb: Good catch! It's fixed

AnalyticsTests/AnalyticsTests.swift
fathyb: I added unit tests with nested structures

Analytics/Classes/SEGAnalyticsConfiguration.h
```diff
@@ -140,4 +140,30 @@ typedef NSMutableURLRequest *_Nonnull (^SEGRequestFactory)(NSURL *_Nonnull);
...
+ * @endcode
+ *
+ * This will replace any matching occurences to a redacted version:
+ * @code
+ * "fb123456789://authorize#access_token=secretsecretsecretsecret&some=data"
+ * @endcode
+ *
+ * Becomes:
+ * @code
+ * "fb123456789://authorize#access_token=((redacted/fb-auth-token))"
```
f2prateek: given the example `fb123456789://authorize#access_token=secretsecretsecretsecret&some=data`, wouldn't the desired output be `fb123456789://authorize#access_token=((redacted/fb-auth-token))&some=data`? (i.e. `&some=data`  at the end.

gerhard-segment:
Good point.

What is the risk of adding this ?

f2prateek:
I just realized there is logging already - https://github.com/segmentio/analytics-ios/pull/798/files#diff-ecadcecb2e3287f62402c496069b01a2R85. This looks reasonable to me!

Analytics/Classes/SEGAnalyticsConfiguration.h
fathyb: Facebook sends other private query parameters to the deep link that could be considered sensible, so we remove everything until we get a space (`[^ ]`).

f2prateek: Got it - sounds good 👍

PR #800, opened 2018-11-29
Remove stray log

f2prateek:


f2prateek:
Noticed this when reading the code. Probably was left in from debugging.

PR #804, opened 2019-01-24
Update topViewController to handle UITabBarController

SBourgon:
(I did  check to see about merging against dev, but there is no dev branch currently)

**What does this PR do?**
Updated seg_topViewController:(UIViewController *) rootViewController to handle the case where the root view controller is a UITabBarController and not only a UINavigationController. 

Auto-tracked events are on with UITabBarController as rootViewController

As our rootViewController was a subclass UITabBarController every auto-tracked event came back as Optimity.Main (for MainViewController.swift)


bsneed:
cc @migs647

migs647:
This was indirectly fixed with PR: https://github.com/segmentio/analytics-ios/pull/899

PR #805, opened 2019-02-08
Removed CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF warnings

adimango:
**What does this PR do?**

CocoaPods enabled `CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF` by
default, so enabled it in the project and update the code accordingly.

Blocks within the library implementation direct touch ivars back properties
as part of updating state with doing the work the public access to the
property requires, so use `self->` notation to make it an explicit
capture of `self` instead of an implicit one.

CocoaPods `CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF`


bsneed:
cc @migs647

migs647:
This was repaired with PR: https://github.com/segmentio/analytics-ios/pull/815

PR #812, opened 2019-03-12
Remove broken .clang-format symlink

tallpants:
**What does this PR do?**
Removes the broken `.clang-format` symlink.

Nope, but you might need to commit a real `.clang-format` file.

 It was a symlink to a file on one of the dev's local machines, so there's really no reason at all to have it here. If you insist on having a `.clang-format` file in the repo we ought to commit a real one.

No tickets.

**Questions:**
- Does the docs need an update? -- No
- Are there any security concerns? -- No
- Do we need to update engineering / success? -- No

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,1 +1,0 @@
-/Users/prateek/dev/src/github.com/square/spacecommander/.clang-format


f2prateek:
This seems to break the CI build https://circleci.com/gh/segmentio/analytics-ios/5690

tallpants:
@f2prateek could you commit the real `.clang-format` file to this branch? I think you have permission to push commits to it since you're a maintainer.

tallpants:
I think that should fix the CI build.

f2prateek:
Looks like the CI is still broken?

For reference, here's the file I had used https://github.com/square/spacecommander/blob/master/.clang-format

f2prateek:
Seems like maybe we should use it via a submodule (https://github.com/square/spacecommander#details)

PR #813, opened 2019-03-13
add stale configuration

f2prateek:
This PR adds a config for [stale](https://probot.github.io/apps/stale/) to clean up our old issues and PRs.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,12 @@
+daysUntilStale: 60
+daysUntilClose: 7
+exemptLabels:
+  - pinned
+  - security
+staleLabel: wontfix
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,12 @@
+daysUntilStale: 60
+daysUntilClose: 7
```
nhi-nguyen: what's our general policy? if there is none, perhaps we should have one written up to be implemented across all libraries?

f2prateek: starting one here https://paper.dropbox.com/doc/Libraries-Stale-Policy--AZW~MXAQ_Pn~fUKEZhq9vYbYAg-tiUzSGiTErmeXLWEdXNbW

PR #822, opened 2019-05-16
Fix app updated event

mihai8804858:
**What does this PR do?**
Fixes an issue when `Application Updated` event is not being triggered under some circumstances.

If you have a test application, try open the app under these conditions:
* There is no info stored about previous version / build -> `Application Installed` should be triggered
* There is info stored about previous version / build
    * Previous and current versions are the same
        * Previous and current builds the same -> No event should be triggered
        * Previous and current builds are different -> `Application Updated` should be triggered
    * Previous and current versions are different
        * Previous and current builds the same -> `Application Updated` should be triggered
        * Previous and current builds are different -> `Application Updated` should be triggered

The current implementation id `Application Updated` event only checks if previous and current builds are different. This means that if version has changed but builds didn't, this event will not be triggered.

**Questions:**
- Does the docs need an update? - NO
- Are there any security concerns? - NO
- Do we need to update engineering / success? - NO

### DIFF
--- a/Analytics/Classes/SEGAnalytics.m
+++ b/Analytics/Classes/SEGAnalytics.m
@@ -138,13 +138,17 @@ NSString *const SEGBuildKeyV2 = @"SEGBuildKeyV2";
             @"version" : currentVersion ?: @"",
             @"build" : currentBuild ?: @"",
         }];
-    } else if (![currentBuild isEqualToString:previousBuildV2]) {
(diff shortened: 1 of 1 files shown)


bsneed:
Closing this since it's so out of date.  Please reopen if it's brought to current. Thanks for the contribution and my sincere apologies it's taken so long to be merged.

PR #825, opened 2019-06-05
Support UIKit for Mac apps

fathyb:


bsneed:
cc @migs647

mattgallagher:
This might be the wrong approach – trying to subtract macOS Catalyst from iOS targets. Instead, since CoreTelephony.framework exists on macOS Catalyst, it might be better to simply link it on macOS since the real problem I suspect this PR is aiming to fix is that CoreTelephony is not automatically linked for macOS Catalyst.

For example, I was getting:

```
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_CTTelephonyNetworkInfo", referenced from:
      objc-class-ref in Analytics.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

but adding the following to the Package.swift file after the cSettings:

```
            linkerSettings: [
                .linkedFramework("CoreTelephony", .when(platforms: [.iOS]))
            ]
```

fixes the problem.

bsneed:
This has been done in 4.0.4.

PR #831, opened 2019-07-16
Add iOS Backgrounded Event.

carlos-kelly:
**What does this PR do?**
Adds the iOS Backgrounded event to the tracking for lifecycle events when enabled

Link against the library and send the app to background. You should see the backgrounded event.

Link against the library and send the app to background. You should see the backgrounded event.

This is part of the work from Formidable 

LIB-1143

**Questions:**
- Does the docs need an update? No (The docs already mention the event)
- Do we need to update engineering / success? No

### DIFF
--- a/Analytics/Classes/SEGAnalytics.m
+++ b/Analytics/Classes/SEGAnalytics.m
@@ -179,7 +179,7 @@ NSString *const SEGBuildKeyV2 = @"SEGBuildKeyV2";
   if (!self.configuration.trackApplicationLifecycleEvents) {
     return;
   }
-  [self track: @"Application Backgrounded" properties: @{}];
+  [self track: @"Application Backgrounded"];
 }
 
 
--- a/Analytics/Classes/SEGAnalytics.m
+++ b/Analytics/Classes/SEGAnalytics.m
@@ -169,12 +169,8 @@ NSString *const SEGBuildKeyV2 = @"SEGBuildKeyV2";
     if (!self.configuration.trackApplicationLifecycleEvents) {
         return;
     }
-    NSString *currentVersion = [[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"];
-    NSString *currentBuild = [[NSBundle mainBundle] infoDictionary][@"CFBundleVersion"];
     [self track:@"Application Opened" properties:@{
         @"from_background" : @YES,
-        @"version" : currentVersion ?: @"",
-        @"build" : currentBuild ?: @"",
     }];
 }
 
@@ -183,13 +179,7 @@ NSString *const SEGBuildKeyV2 = @"SEGBuildKeyV2";
   if (!self.configuration.trackApplicationLifecycleEvents) {
     return;
   }
-  NSString *currentVersion = [[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"];
-  NSString *currentBuild = [[NSBundle mainBundle] infoDictionary][@"CFBundleVersion"];
-  [self track: @"Application Backgrounded" properties: @{
-    @"from_background" : @NO,
-    @"version" : currentVersion ?: @"",
-    @"build" : currentBuild ?: @"",
-  }];
+  [self track: @"Application Backgrounded" properties: @{}];
 }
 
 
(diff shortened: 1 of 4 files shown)


Analytics/Classes/SEGAnalytics.m
```diff
@@ -111,6 +111,8 @@ - (void)handleAppStateNotification:(NSNotification *)note
         [self _applicationDidFinishLaunchingWithOptions:note.userInfo];
     } else if ([note.name isEqualToString:UIApplicationWillEnterForegroundNotification]) {
         [self _applicationWillEnterForeground];
+    } else if ([note.name isEqualToString: UIApplicationDidEnterBackgroundNotification]) {
```
f2prateek: I believe this would be fired even if the application is being suspended, which is not the behavior that is desired in the PRD.

However, I don't believe it's possible to accurately capture the behavior documented in the PRD, and this code as implemented is the right solution, and we should update the PRD instead.

In case the PRD is not updated, then we might have to update the code here.

Analytics/Classes/SEGAnalytics.m
```diff
@@ -176,6 +178,20 @@ - (void)_applicationWillEnterForeground
...
 }
 
+- (void)_applicationDidEnterBackground
+{
+  if (!self.configuration.trackApplicationLifecycleEvents) {
+    return;
+  }
+  NSString *currentVersion = [[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"];
+  NSString *currentBuild = [[NSBundle mainBundle] infoDictionary][@"CFBundleVersion"];
+  [self track: @"Application Backgrounded" properties: @{
```
f2prateek: I don't believe any properties should be collected here, as per https://segment.com/docs/spec/mobile/#application-backgrounded.

Analytics/Classes/SEGAnalytics.m
f2prateek: Agreed to update the PRD - so this part is 👍

f2prateek:
cc @fathyb @bsneed

Analytics/Classes/SEGAnalytics.m
```diff
@@ -183,12 +179,8 @@ - (void)_applicationDidEnterBackground
   if (!self.configuration.trackApplicationLifecycleEvents) {
     return;
   }
-  NSString *currentVersion = [[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"];
-  NSString *currentBuild = [[NSBundle mainBundle] infoDictionary][@"CFBundleVersion"];
   [self track: @"Application Backgrounded" properties: @{
-    @"from_background" : @NO,
-    @"version" : currentVersion ?: @"",
-    @"build" : currentBuild ?: @"",
+    @"from_background" : @NO
```
f2prateek: This should be removed as well.

Analytics/Classes/SEGAnalytics.m
```diff
@@ -169,12 +169,8 @@ - (void)_applicationWillEnterForeground
     if (!self.configuration.trackApplicationLifecycleEvents) {
         return;
     }
-    NSString *currentVersion = [[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"];
-    NSString *currentBuild = [[NSBundle mainBundle] infoDictionary][@"CFBundleVersion"];
     [self track:@"Application Opened" properties:@{
-        @"from_background" : @YES,
-        @"version" : currentVersion ?: @"",
-        @"build" : currentBuild ?: @"",
+        @"from_background" : @YES
```
f2prateek: This should still collect the properties outlined in https://segment.com/docs/spec/mobile/#application-opened

Analytics/Classes/SEGAnalytics.m
```diff
@@ -167,15 +169,19 @@ - (void)_applicationWillEnterForeground
...
-        @"build" : currentBuild ?: @"",
     }];
 }
 
+- (void)_applicationDidEnterBackground
+{
+  if (!self.configuration.trackApplicationLifecycleEvents) {
+    return;
+  }
+  [self track: @"Application Backgrounded" properties: @{}];
```
f2prateek: `[self track: @"Application Backgrounded"];` should also suffice here.

Analytics/Classes/SEGAnalytics.m
bsneed: Fwiw, you're going to get applicationDidEnterBackground when the device is powered off and the app is still active, as well as if they force-quit from the task switcher.  If we're cool with that, then approved.

PR #833, opened 2019-07-19
adding “name” field to payload;updated nimble to version 7.3.4

danieljackins:
**What does this PR do?**
Add's collection of "name" field to the iOS library, and updates Nimble to version 7.3.4
N/A
N/A
Done to match collection of data in Android library. Android was collecting both the model and name of the device, whereas iOS was only collecting the model. Libraries will now have parity and match the Segment Spec.
https://segment.atlassian.net/browse/LIB-837
N/A
**Questions:**
No, this is how functionality was outlined in the Spec to begin with.
N/A
Unsure

### DIFF
--- a/Analytics.xcodeproj/project.pbxproj
+++ b/Analytics.xcodeproj/project.pbxproj
@@ -541,7 +541,7 @@
 			files = (
 			);
 			inputPaths = (
-				"${SRCROOT}/Pods/Target Support Files/Pods-AnalyticsTests/Pods-AnalyticsTests-frameworks.sh",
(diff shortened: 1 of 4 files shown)


f2prateek:
I'd generally like to see unrelated changes split into different PRs. So in this case, I'd do one for the Nimble update and one for the `name` change.

For the `name` change - is there a test we could add as well?

bsneed:
> I'd generally like to see unrelated changes split into different PRs. So in this case, I'd do one for the Nimble update and one for the `name` change.


In general, I'd agree, however in this case I feel like it adds unnecessary work since that change was required just to get tests to compile.

Re the test, if we can add it quickly, we should.  If not, which is what I suspect given our tests can't be run in isolation, maybe skip it for now since this is overall a one-line change.

PR #839, opened 2019-07-26
Add support for SSL Pinning

ujjawalgarg:
**What does this PR do?**
This PR provides scaffolding for client integrations to implement SSL pinning checks if required

Client integrations can optionally pass in a NSURLSessionDelegate in the SEGAnalyticsConfiguration object.

If set, NSURLSessionDelegate callbacks are forwarded to the client code where SSL pinning checks can be implemented

AnalyticsTests has been modified to test httpSessionDelegate configuration

Currently, Segment SDK does not provide SSL pinning support. We want to enable this.

**Questions:**
yes

no

yes


bsneed:
Verified in a test project. This will be released a little later as we're queuing up a number of changes.

Issue #841, opened 2019-08-19
CoreTelephony is not part of tvOS -- getting link errors


skydivedan:
CoreTelephony is not part of tvOS. 

Could you modify the podspec so that it doesn't include `CoreTelephony` for tvOS targets? Right now I'm getting linker errors.

Zanchee:
Any update on when this will be released, we would like to update to 3.7 but can't. until this fix is released.

bsneed:
It was released in 3.8.0-beta.1.  👍

PR #842, opened 2019-08-19
`CoreTelephony` library is now only included on iOS targets. 

skydivedan:
**What does this PR do?**
This makes sure that `CoreTelephony` is only included for iOS targets.

Build a tvOS target with Cocoapods. Include `pod 'Analytics'` as a dependency. Building for actual tvOS hardware (not simulator) will fail. Using this changed podspec will remove the dependency for `CoreTelephony` in tvOS, and will fix the linker error.

https://github.com/segmentio/analytics-ios/issues/841

**Questions:**
  No
  No
  No

### DIFF
--- a/Analytics.podspec
+++ b/Analytics.podspec
@@ -17,7 +17,8 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = '7.0'
   s.tvos.deployment_target = '9.0'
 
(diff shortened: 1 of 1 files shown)


bsneed:
Verified in a test project.  This will be released a little later as we're queuing up a number of changes.

Zanchee:
Any news on when 3.8 will be released from beta? We are unable to currently update  Segment due to this issue.

Issue #824, opened 2019-06-04
Support for Swift Package Manager


WarSunsOfPeace:
With iOS support add for Swift Package Manager in Xcode 11 can we get that added to this project.

rosskimes:
Any plans for this now that Xcode 11 (with SwiftPM support) is officially out?

sourabhguptavc:
Any update on this issue?

bdrelling:
Any updates?

bsneed:
I left some commentary on your PR linked above.  This will likely not make it in until the 4.0 beta arrives toward March, but I'll try to make sure it's included there.

bdrelling:
@bsneed Wondering if there's any more updates on this.

Also, I am not the author of any PRs opened in this repository. I believe you left a comment for another author that seems to have gone MIA, unfortunately.

bsneed:
@bdrelling it is on our roadmap, yes.  While I don't have an exact date just yet for it's inclusion other than soon, it will definitely make it in along with swift language improvements.

andreyz:
+1 on SwiftPM support. What's stopping you?

ninjaproger:
+1

bsneed:
To expedite this inclusion, a PR that has working test suite, and verifies that Cocoapods and Carthage continue to work will be more helpful than +1's.  I've asked these questions a number of times but no one seems to answer.  I thank you all for your interest, but in terms of priorities, this is low beyond my own personal wish that it get done.

andreyz:
> To expedite this inclusion, a PR that has working test suite, and verifies that Cocoapods and Carthage continue to work will be more helpful than +1's. I've asked these questions a number of times but no one seems to answer. I thank you all for your interest, but in terms of priorities, this is low beyond my own personal wish that it get done.

Well, I don't get paid to support a paid SaaS SDK, but someone developing this library does. I did gave a conversion a quick stab, but this was much more involved than 1h I was willing to invest in already paid product (Segment).

bsneed:
@andreyz thanks for your attempt.  if you can't make the changes and this is critical to you, please  contact Segment support and escalate through there.  That's part of what drives our priorities.  As I mentioned, this is on our roadmap because *we*, the libraries team here at Segment, want it.  My apologies that the timing of when we'll be able to do it, and your desire for it don't line up.

dmiluski:
[...]

PR #845, opened 2019-09-25
Added support for Swift Package Manager

mark-shape:
Just took a second look and reduced the code change a bit for my previous pull request. The conclusion I came to, was that when Swift imports the headers, all the `#import` needs to point to location on disk, as the "Xcode magic" isn't used by SwiftPM. This will require going forward, that the imports are written with the relative path. The `headerSearchPath` in `Package.swift` does allow implementation files to continue just using the header filename in imports.

**What does this PR do?**

Adds support for Swift Package Manager

By adding the branch as a dependency to a test app in Xcode 11

There should be no functional changes, so just the basic "see if it works" testing after the library has been added to the Xcode project

We are in the progress of swapping out Carthage with Swift Package Manager and was only missing this library 🙂 

We have not opened any tickets, but is related to #824 

No UI changes

**Questions:**

Maybe just to indicate that the library can be installed with SwiftPM

I don't see there should be any as there are no functional changes to the code.

### DIFF
--- a/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+++ b/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "self:">
+   </FileRef>
+</Workspace>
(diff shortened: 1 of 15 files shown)


bdrelling:
@bsneed just pinging a contributor--any thoughts around this or #824? Thanks!

bsneed:
I'll see if I can get a thorough review scheduled shortly.  I see some header locations have changed, did you verify that it still builds as a cocoa pod?

Issue #846, opened 2019-09-26
SEGFileStorage dataFromPlist:


leandromperez:
Hi, 
We're seeing several crashes with the following stack trace:
```
#0	(null) in CFGetTypeID ()
#1	(null) in _CFAppendXML0 ()
#2	(null) in _CFAppendXML0 ()
#3	(null) in _CFAppendXML0 ()
#4	(null) in _CFAppendXML0 ()
#5	(null) in _CFAppendXML0 ()
#6	(null) in _CFPropertyListCreateXMLData ()
#7	(null) in CFPropertyListCreateData ()
#8	(null) in +[NSPropertyListSerialization dataWithPropertyList:format:options:error:] ()
#9	0x0000000105539cb8 in -[SEGFileStorage dataFromPlist:] at /MyApp/Pods/Analytics/Analytics/Classes/Internal/SEGFileStorage.m:149
#10	0x0000000105539c28 in -[SEGFileStorage setPlist:forKey:] at /MyApp/Pods/Analytics/Analytics/Classes/Internal/SEGFileStorage.m:140
#11	0x00000001055399f8 in -[SEGFileStorage setArray:forKey:] at /MyApp/Pods/Analytics/Analytics/Classes/Internal/SEGFileStorage.m:105
#12	0x0000000105544838 in -[SEGSegmentIntegration persistQueue] at /MyApp/Pods/Analytics/Analytics/Classes/Internal/SEGSegmentIntegration.m:630
#13	0x0000000105543964 in -[SEGSegmentIntegration queuePayload:] at /MyApp/Pods/Analytics/Analytics/Classes/Internal/SEGSegmentIntegration.m:470
#14	0x00000001055437b4 in __71-[SEGSegmentIntegration enqueueAction:dictionary:context:integrations:]_block_invoke at /MyApp/Pods/Analytics/Analytics/Classes/Internal/SEGSegmentIntegration.m:460
#15	0x00000001055386bc in __seg_dispatch_specific_block_invoke at /MyApp/Pods/Analytics/Analytics/Classes/Internal/SEGAnalyticsUtils.m:64
#16	(null) in _dispatch_call_block_and_release ()
#17	(null) in _dispatch_client_callout ()
#18	(null) in _dispatch_lane_serial_drain$VARIANT$armv81 ()
#19	(null) in _dispatch_lane_invoke$VARIANT$armv81 ()
#20	(null) in _dispatch_workloop_worker_thread ()
[...]

PR #853, opened 2019-11-19
Fix for LIB-1416 & Github #846

bsneed:
- Stops blindly passing dictionaries.
- Property dictionaries are checked for NSCoding conformance to ensure they can be serialized.
- Non NSCoding conformant properties trigger an assert in DEBUG builds.
- Property dictionaries are deep copied so contents can’t change while the pipeline is in progress.
- Puts a try/catch arrangement as a temporary guard against crashes for serialization failures until the storage format can be changed.


bsneed:
> Should we test this?

Ah thanks for catching.  They're in the other follow on PR I'm building.  One sec.

Issue #852, opened 2019-11-13
iOS Segment.io SDK does not allow `null` in Event's properties


abodnyaUA:
We use the SDK to send analytics on Segment.io by calling
```[SEGAnalytics - (void)track:(NSString *)event properties:(SERIALIZABLE_DICT _Nullable)properties]``` method. This method uses `SEGCoerceJSONObject` method to serialize properties in JSON and during the serialization, it drops all the NSNull values.
Because of this we cannot send as a `properties` a dictionary with the value for key equals null, for example `{ "name": null }`

Android's implementation of Segment.io SDK **ALLOWS** sending null values.

In the comment of SEGCoerceJSONObject method it says that dropping NSNull is a HOTFIX
```// Hotfix for issue where SEGFileStorage uses plist which does NOT support NSNull```
This "HOTFIX" was added more than 2 years ago in the pr https://github.com/segmentio/analytics-ios/pull/707

Could you please provide normal fix for the issue and stop dropping NSNull values?
The iOS SDK is inconsistent with Android one and behave differently.

Thanks!

abodnyaUA:
cc @tonyxiao @ladanazita @f2prateek @TeresaNesteby @myclamm

bsneed:
Switched the underlying storage from plist to json.  Null's are now coming across correctly.  PR to be attached shortly.

abodnyaUA:
@bsneed, wait for the pull-request. Thanks!

bsneed:
Fixed, but leaving this open until we roll out another release soon.

bsneed:
Released in 3.8.0-beta.1

PR #854, opened 2019-11-26
Swapped JSON in for the storage format instead of plists.

bsneed:
**What does this PR do?**

- Changes the storage format for SEGFileStorage to use JSON instead of plists.
- Perform conversion on existing plist format files as encountered.

LIB-1494 and #852

**Questions:**
NO
NO
NO

### DIFF
--- a/Analytics/Classes/Internal/SEGFileStorage.m
+++ b/Analytics/Classes/Internal/SEGFileStorage.m
@@ -107,12 +107,18 @@
 
 - (NSString *)stringForKey:(NSString *)key
 {
-    return [self jsonForKey:key];
+    // unlike plists, we can't have stray values, they need to be
+    // in a dictionary or array container.
+    NSDictionary *data = [self jsonForKey:key];
+    return data[key];
 }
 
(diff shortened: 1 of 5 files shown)


Analytics/Classes/Internal/SEGFileStorage.m
```diff
@@ -129,20 +135,80 @@ - (NSURL *)urlForKey:(NSString *)key
...
+- (id _Nullable)jsonForKey:(NSString *)key
 {
+    id result = nil;
+    
     NSData *data = [self dataForKey:key];
-    return data ? [self plistFromData:data] : nil;
+    if (data) {
+        BOOL needsConversion = NO;
+        result = [self jsonFromData:data needsConversion:&needsConversion];
+        if (needsConversion) {
```
danieljackins: isn't needsConversion always going to be false here?

bsneed: No, the jsonFromData:needsConversion: method will set it's value if necessary.

PR #859, opened 2020-01-24
LIB-1556: Source Middleware refactor for Edge Functions

bsneed:
**What does this PR do?**

- Refactors source middleware.
- Fixes compiler warnings.
- Adds new test suite for source middleware not based on quick/nimble.
- Fixes bugs encountered along the way.
- Moves iOS compatibility to 12.x+ 

SDD here: https://paper.dropbox.com/doc/SDD-Edge-Functions-Phase-2--AsXj~NjVk5xmJFg1JRDdmjzsAg-DbpY8qNkdg2ybhX9UcpCg


Analytics/Classes/Middlewares/SEGContext.m
```diff
@@ -68,10 +68,10 @@ - (id)copyWithZone:(NSZone *)zone
 {
     SEGContext *ctx = [[SEGContext allocWithZone:zone] initWithAnalytics:self._analytics];
     ctx.eventType = self.eventType;
-    ctx.userId = self.userId;
-    ctx.anonymousId = self.anonymousId;
-    ctx.payload = self.payload;
-    ctx.error = self.error;
+    ctx.userId = [self.userId copy];
+    ctx.anonymousId = [self.anonymousId copy];
+    ctx.payload = self.payload; // FIX: this needs to be copied!!
```
fathyb: What happens if we don't copy it?

Analytics/Classes/Integrations/SEGIdentifyPayload.h
```diff
@@ -3,7 +3,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-
+NS_SWIFT_NAME(IdentifyPayload)
```
fathyb: Is this a breaking change?

Analytics/Classes/Middlewares/SEGContext.m
bsneed: Nothing, *yet*.  For destination middleware this would be problematic the same payload dictionary would be getting modified all over.

Analytics/Classes/Integrations/SEGIdentifyPayload.h
bsneed: Yes.

PR #865, opened 2020-02-20
Publish filenames used for data storage

bejo:
**What does this PR do?**
Makes filenames used for data storage public.

It's very short.

I don't think there's anything to test.

Yes, so I need those filenames to be public since my application often starts before a device is unlocked. In such cases Segment library is booted but it fails to read its files/data and then reports analytic events without a `user_id`. Moreover, during such "corrupted" sessions it's using a new `anonymous_id`.

A followup-fix on my application side will be to change `FileProtectionType` of the library files to `.none`. I'm not sure if that's something you'd like to do for all your users, so this PR only publishes those file names.

n/a

n/a

**Questions:**
- Does the docs need an update? I added docs for the `const`s which I made public.
- Are there any security concerns? No.
- Do we need to update engineering / success? TBH, I don't understand the question. That said, you guys might want to do two things:
1) Group all your files in `com.segment.analytics` directory within `Application Support` directory. Just to clean up / separate things a bit.
2) Set `FileProtectionType` of all your files to `.none` or... figure out which files are needed during boot-time and which can be accessed later (when a device is unlocked).

### DIFF
--- a/Analytics/Classes/Integrations/SEGIntegrationsManager.m
+++ b/Analytics/Classes/Integrations/SEGIntegrationsManager.m
@@ -26,9 +26,9 @@
 #import "SEGAliasPayload.h"
 
 NSString *SEGAnalyticsIntegrationDidStart = @"io.segment.analytics.integration.did.start";
-static NSString *const SEGAnonymousIdKey = @"SEGAnonymousId";
-static NSString *const kSEGAnonymousIdFilename = @"segment.anonymousId";
-static NSString *const SEGCachedSettingsKey = @"analytics.settings.v2.plist";
+NSString *const SEGAnonymousIdKey = @"SEGAnonymousId";
+NSString *const kSEGAnonymousIdFilename = @"segment.anonymousId";
+NSString *const kSEGCachedSettingsFilename = @"analytics.settings.v2.plist";
 
 
 @interface SEGAnalyticsConfiguration (Private)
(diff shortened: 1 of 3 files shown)


Analytics/Classes/Integrations/SEGIntegrationsManager.m
```diff
@@ -26,9 +26,9 @@
 #import "SEGAliasPayload.h"
 
 NSString *SEGAnalyticsIntegrationDidStart = @"io.segment.analytics.integration.did.start";
-static NSString *const SEGAnonymousIdKey = @"SEGAnonymousId";
-static NSString *const kSEGAnonymousIdFilename = @"segment.anonymousId";
-static NSString *const SEGCachedSettingsKey = @"analytics.settings.v2.plist";
+NSString *const SEGAnonymousIdKey = @"SEGAnonymousId";
+NSString *const kSEGAnonymousIdFilename = @"segment.anonymousId";
+NSString *const kSEGCachedSettingsFilename = @"analytics.settings.v2.plist";
```
bejo: I also changed this `const` name in order to follow the convention established by other filenames/constants whin the project.

bsneed:
Hi @bejo; Sorry it took me awhile to review this PR.  The changes look fine, however I'm not sure I quite understand your use case. It seems that setting file protection type to None on files that contain user-identifiable information might not be a good idea.

bejo:
> setting file protection type to None on files that contain user-identifiable information might not be a good idea

I totally agree it's not ideal :( That said, if the app starts before the device is unlocked I don't see any other way of getting the library to work properly.

I guess an alternative could be for the library to accumulate events and dispatch them only once it has access to file system. However, that would be a pretty big change 😬 

> I see a number of places where the const changes don't match up anymore.

Ops, looks like I forgot to run the tests for TvOS. Fixed now! ✅

bsneed:
Thanks @bejo!

PR #880, opened 2020-04-23
Fixed a crash in the file storage

arietis:
**What does this PR do?**
Fixed a crash in the file storage when trying to get a string stored using old SDK version.

`SEGFileStorage.m`

You need a version of SDK where the string is converted to Data directly without encapsulating it into the Dictionary object. After that upgrade to the version from the PR and it shouldn't crash.

To avoid such crashes in the future some migration logic would be helpful.

No tickets I know about.

No UI/UX changes.

**Questions:**
No.
No.

No.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ Carthage
 # itself has no pod dependencies, only our tests do.
 Pods/
 .clang-format
+.idea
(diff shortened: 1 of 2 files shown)


bsneed:
@arietis thanks so much for this!!

Issue #869, opened 2020-03-09
registeredForRemoteNotificationsWithDeviceToken doesn't accept FirebaseCloudMessaging token


bteng22:
Our team's notifications are powered through Firebase Cloud Messaging. When `didRegisterForRemoteNotificationsWithDeviceToken` callback is invoked we have a handle to the fcmToken through `Messaging.messaging().fcmToken` and we're forwarding this token through to Segment's `registeredForRemoteNotificationsWithDeviceToken` as shown below

```
func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
  if Messaging.messaging().fcmToken != nil {
    let firebaseCloudMessagingToken = Data(Messaging.messaging().fcmToken!.utf8)
    SEGAnalytics.shared()?.registeredForRemoteNotifications(withDeviceToken: firebaseCloudMessagingToken)
  }
}
```

However, whenever we read the token on Segment, the device ID is invalid. We confirmed the `fcmToken` is correct by logging it locally and successfully sending notifications to the device directly, but somewhere along the way Segment is decoding it differently than we expect. Are we missing something here or does Segment need to be able to accept device tokens from both APNS and FCM?

The code in suspect is here https://github.com/segmentio/analytics-ios/blob/master/Analytics/Classes/Internal/SEGSegmentIntegration.m#L410

bsneed:
Hi @bteng22!  I think what should be getting passed to Segment is the value passed into the deviceToken parameter, not the fcmToken.  I don't know what the fcmToken looks like as a string off hand, but we're expecting to convert it to a hexadecimal string.  If it's not coming across correctly, I suspect this conversion is failing due to the contents of the data it's being given.

nutlike:
Is there another possibility to attach the deviceToken to the device context? I also want to pass the FCM token to Segment (which would relay it to Customer.io which we use for sending push notifications [via Firebase Cloud Messaging]). The Segment Android SDK offers a [putDeviceToken](https://github.com/segmentio/analytics-android/blob/9ad041fbec94424f321758cb73ed2683f3f7f5e5/analytics/src/main/java/com/segment/analytics/AnalyticsContext.java#L250) method for this which adds the information to the static device context. I am not super familiar with iOS/Objective-C and any pointers into the right direction would be appreciated.

levibostian:
Commenting on this issue as it is still relevant today. 

[...]

PR #881, opened 2020-04-28
Add method to put device token

nutlike:
**What does this PR do?**
This PR adds a method to put device tokens which are utilized by some destinations (e.g. Customer.io). [The Segment Android SDK offers the same functionality.](https://github.com/segmentio/analytics-android/blob/9ad041fbec94424f321758cb73ed2683f3f7f5e5/analytics/src/main/java/com/segment/analytics/AnalyticsContext.java#L443-L447) Setting device tokens (obtained from Firebase Cloud Messaging SDK) enables Customer.io to track devices via Segment which enables their push notification feature.

Call `[SEGAnalytics.sharedAnalytics putDeviceToken:@"i-am-a-device-token"]` and check in Segment’s event debugger if new events contain the `token` key within their `context/device` dictionary.

We are developing in React Native and are using Firebase Cloud Messaging (via [React Native Firebase](https://rnfirebase.io/)). We want to use Customer.io to send out push notifications, so we are working on a pull request for the Segment React Native SDK to add functionality which is required to make this happen. Before we can continue our work we need a way to set a FCM device token in iOS – hence this pull request.

#869

**Questions:**
  I added comments, but the Segment website ([Analytics for iOS](https://segment.com/docs/connections/sources/catalog/libraries/mobile/ios/)) might also need one.
  I don't think so.
  Maybe?

### DIFF
--- a/Analytics/Classes/Integrations/SEGIntegration.h
+++ b/Analytics/Classes/Integrations/SEGIntegration.h
@@ -59,6 +59,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)receivedRemoteNotification:(NSDictionary *)userInfo;
 - (void)failedToRegisterForRemoteNotificationsWithError:(NSError *)error;
 - (void)registeredForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken;
+- (void)putDeviceToken:(NSString *)deviceToken;
 - (void)handleActionWithIdentifier:(NSString *)identifier forRemoteNotification:(NSDictionary *)userInfo;
 
 // Callbacks for app state changes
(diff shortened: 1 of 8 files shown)


bteng22:
Thanks for putting up the PR @nutlike 👍  Our team needs the same functionality. Do we just need to resolve these conflicts to get this merged in?

nutlike:
As far as I understood this PR won‘t get merged, see https://github.com/segmentio/analytics-react-native/pull/170#issuecomment-634178415: I still didn‘t find the time to look into the other options mentioned (and I believe there have been quite some changes to the SDKs since I opened this PR) – as soon as I have another solution I will post it here.

harrigee:
We also need this functionality. Till this is fixed we are forced skip segment and use customer.io directly.

bsneed:
Closing this. Further instructions show neither this solution or middleware are necessary.

Simply calling registeredForRemoteNotificationsWithDeviceToken: will do the trick and device.token will be set even if you don't need a full-on push notification setup. Segment doesn't care what the value is, we just throw it in the payload as requested.

PR #912, opened 2020-06-25
Payload Info & Traits Fixes

bsneed:
- Moves the location of top level payload info (anonymousId, messageId, etc) from SEGContext into SEGPayload.
- Fixes issue with traits not being restored correctly after app relaunch.
- Adds tests for said traits issue.


Analytics/Classes/SEGPayload.h
```diff
@@ -10,6 +10,8 @@ NS_SWIFT_NAME(Payload)
 @property (nonatomic, readonly) JSON_DICT integrations;
 @property (nonatomic, strong) NSString *timestamp;
 @property (nonatomic, strong) NSString *messageId;
+@property (nonatomic, strong) NSString *anonymousId;
```
migs647, line 13: Should these next two lines be copy?

Issue #916, opened 2020-06-30
Crash on app startup with Segment 4.0.1


jbelkins:
The Analytics SDK is crashing my app on startup, starting with version 4.  (Crashes on 4.0.0 and 4.0.1)
Everything ran just fine for us on version 3.

A representative crash log from my crash reporter and a Xcode screenshot showing the originating thread are attached.

It appears that you're enqueueing a synchronous task from within a task on the same queue, hence: deadlock.  The point of entry to the Analytics SDK was when we called `registeredForRemoteNotifications(withDeviceToken:)`

[Crash log](https://github.com/segmentio/analytics-ios/files/4854750/bugsnag_error_stacktrace_EXC_BREAKPOINT_event_5ef4b14d005b713de55f0000.txt)
[Xcode screen shot](https://user-images.githubusercontent.com/863266/86185279-95bd7300-bafb-11ea-81a7-ce0efebc1e96.png)

bsneed:
@jbelkins thanks for reporting this!  We'll get this fixed ASAP.

bsneed:
Fixed on master.  Release coming soon.

PR #920, opened 2020-07-13
Miscellaneous Fixes

bsneed:
- Fixes issue #916, Deadlock when setting device token. 
- Adopts Swift 5 syntax in tests.
- Removed SwiftTryCatch dependency in tests.
- Fixed issue where userId was not being cleared properly on reset.


AnalyticsTests/AnalyticsTests.swift
```diff
@@ -259,4 +259,21 @@ class AnalyticsTests: XCTestCase {
         integration?.test_fileStorage()?.resetAll()
         XCTAssert(integration?.test_queue()?.isEmpty ?? false)
     }
+    
+    func testDeviceTokenRegistration() {
+        func getStringFrom(token: Data) -> String {
+            return token.reduce("") { $0 + String(format: "%02.2hhx", $1) }
```
migs647, line 265: Might be nice to have a comment on this line to explain the format that is being converted

Analytics/Internal/SEGState.m
```diff
@@ -161,8 +161,8 @@ - (void)setDeviceToken:(NSString *)deviceToken
 {
     [state setValueWithBlock: ^{
         self->_deviceToken = [deviceToken copy];
-        [self updateStaticContext];
     }];
+    [self updateStaticContext];
```
migs647, line 165: Great catch!

Issue #936, opened 2020-08-24
Crash SEGFileStorage.m - Analytics v4.0.4


vpetrosky:
Since updating to Analytics version 4.0.4, we have a new trending crash. 

An exception is thrown on app start up, when we fire our first few Segment events. It's reported in Firebase as a fatal crash. In my testing, I've been able to hit an exception breakpoint but am able to continue with program execution when debugging. I worry though, that this exception may sometimes result in a fatal crash that's affecting users. 

**Details**
Operating Systems: 100% iOS 13
Device States: 75% background

```Fatal Exception: NSInvalidArgumentException
Invalid type in JSON write (__NSTaggedDate)

-[SEGFileStorage dataFromJSON:]
```

```
Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x1b57c5654 __exceptionPreprocess
1  libobjc.A.dylib                0x1b54e7bcc objc_exception_throw
2  Foundation                     0x1b5c3db7c _writeJSONValue
3  Foundation                     0x1b5c404a0 ___writeJSONObject_block_invoke
4  CoreFoundation                 0x1b5812c14 __NSDICTIONARY_IS_CALLING_OUT_TO_A_BLOCK__
5  CoreFoundation                 0x1b569efc4 -[__NSFrozenDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:]
6  Foundation                     0x1b5c3f9d0 _writeJSONObject
7  Foundation                     0x1b5c404a0 ___writeJSONObject_block_invoke
8  CoreFoundation                 0x1b5812c14 __NSDICTIONARY_IS_CALLING_OUT_TO_A_BLOCK__
9  CoreFoundation                 0x1b569efc4 -[__NSFrozenDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:]
10 Foundation                     0x1b5c3f9d0 _writeJSONObject
11 Foundation                     0x1b5c404a0 ___writeJSONObject_block_invoke
12 CoreFoundation                 0x1b5812c14 __NSDICTIONARY_IS_CALLING_OUT_TO_A_BLOCK__
13 CoreFoundation                 0x1b569efc4 -[__NSFrozenDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:]
14 Foundation                     0x1b5c3f9d0 _writeJSONObject
15 Foundation                     0x1b5c407a8 ___writeJSONArray_block_invoke
[...]

Issue #929, opened 2020-08-12
Crash when calling SEGSegmentIntegration enqueueAction:dicitionary:context:integrations:]


misterkerry:
## Problem
Though it is rare, it seems that occasionally, an attempt to send a message fails causing a crash for our app.

## Screenshot of crash in Organizer

![Screen Shot 2020-08-12 at 2 01 45 PM](https://user-images.githubusercontent.com/62082137/90018163-6a09dd00-dca4-11ea-9095-85492777588c.png)

## Logs
`Hardware Model:      iPhone12,5   
Version:             8 (2.4.1)
AppStoreTools:       11E707
AppVariant:          1:iPhone12,5:13
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1] 

Date/Time:           2020-07-29 09:42:47.9834 -0400
Launch Time:         2020-07-29 08:45:05.1810 -0400
OS Version:          iPhone OS 13.5.1 (17F80)
Release Type:        User
Baseband Version:    1.06.00
Report Version:      104

Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x00000006e8caf270
VM Region Info: 0x6e8caf270 is not in any region.  Bytes after previous region: 18401129073  
      REGION TYPE                      START - END             [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      MALLOC_NANO            0000000280000000-00000002a0000000 [512.0M] rw-/rwx SM=PRV  
--->  
      UNUSED SPACE AT END

Termination Signal: Segmentation fault: 11
Termination Reason: Namespace SIGNAL, Code 0xb
Terminating Process: exc handler [4400]
Triggered by Thread:  27

Thread 0 name:
Thread 0:
0   QuartzCore                    	0x00000001a5c0ae28 hash_table_modify(x_hash_table_struct*, void*, void*, int) + 120 (x-hash.cpp:298)
1   QuartzCore                    	0x00000001a5c0adec hash_table_modify(x_hash_table_struct*, void*, void*, int) + 60 (x-hash.cpp:296)
2   QuartzCore                    	0x00000001a5d1c01c CA::Render::Encoder::encode_object(CA::Render::Object const*) + 344 (x-hash.cpp:346)
[...]

PR #937, opened 2020-08-24
Fixed issue where non-serializable types would get into payload

bsneed:
Fixes issue #936 

- NSCoding conformance check was too broad.  Reduced to explicitly allowed types per Apple documentation.
- Added tests to validate non-serializable types trigger asserts in debug mode.
- Added method to do actual type validation.


rogerluan:
Does this PR fix https://github.com/segmentio/analytics-ios/issues/929?

bsneed:
@rogerluan it might, but given lack of repro steps on #929 I can't say for sure.

tibr:
@bsneed this change removed the ability to serialize `NSURL` objects. I guess that wasn't on purpose given that it killed the [tracking of deep links](https://segment.com/docs/connections/spec/mobile/#application-opened)?

bsneed:
@tibr I'm perplexed as to how this was working before tbh.  NSURL isn't a supported type w/ NSJSONSerializer.  A quick test confirms this.  I'll open an issue for this internally.

tibr:
@bsneed true, but it does conform to `NSCoding` so it wasn't removed from the dictionary in `serializableDeepCopy` and then fixed in [`SEGCoerceJSONObject`](https://github.com/segmentio/analytics-ios/blob/594d0499f12a13db1eab3832030e27ff17f0dff8/Analytics/Internal/SEGUtils.m#L553).

bsneed:
@tibr ah, good find!  I should have a fix pushed to master shortly.  Might be a bit before a release though.

bsneed:
@tibr fix is up on master! :highfive:

PR #950, opened 2020-10-15
Fix configuration setting for edgefn

bsneed:
- Removes configuration shimmy from SEGAnalytics and puts it in SEGAnalyticsConfiguration where it belongs.
- Stops trying to double wrap options in array.


migs647 (approved):
Where the tests




LGTM

Issue #947, opened 2020-10-08
NSAssert: Key `url` is a NSURL and can't be serialized for delivery


krin-san:
Segment cannot handle app start options dictionary on iOS 14 when it contains deeplink url (when application was started to handle that deep link).

Related to #937 and #936
Affected version: [v4.0.5](https://github.com/segmentio/analytics-ios/releases/tag/4.0.5) and 7a8387b1048915b22bbc971ff308baa6a0e7e13a (latest master at the moment of testing).
iOS version: 14.0.1
IDE: Xcode 12.0.1

Issues:
- Assertion in `-[NSDictionary(SerializableDeepCopy) serializableDeepCopy:]` on DEBUG builds
- Missing tracking data (deep link) on other builds

Reasons:
- `NSURL` is not mentioned in https://github.com/segmentio/analytics-ios/blob/7a8387b1048915b22bbc971ff308baa6a0e7e13a/Segment/Internal/SEGUtils.m#L507
- `NSURL` don't conform to `SEGSerializable`

<img width="537" alt="Screenshot 2020-10-08 at 09 30 27" src="https://user-images.githubusercontent.com/2318002/95430602-471a4300-094c-11eb-9fb0-1c4de8eb136b.png">

LLDB log at selected stack trace point:
```
po note.userInfo
{
    UIApplicationLaunchOptionsURLKey = "my_app://deeplink";
}

po [note.userInfo[@"UIApplicationLaunchOptionsURLKey"] class]
NSURL
```

krin-san:
Adding `[obj isKindOfClass:[NSURL class]] ||` to the list of serializable classes seem to fix this issue.

dannyhertz:
Im running into this issue as well. Were you able to patch it locally @krin-san or did you have to fork the library and fix it yourself?

krin-san:
@dannyhertz I patched SEGUtils.m in place, right in Pods folder to see if that's enough. On Monday I can prepare a PR if no one wants to make that earlier.

krin-san:
@dannyhertz it took some time to prioritize and test that, but PR is ready.
Could you check if that fixes your issue too? To install the fork change your Podfile like that:
```diff
- pod 'Analytics'
+ pod 'Analytics', :git => 'https://github.com/krin-san/analytics-ios.git', :branch => 'bugfix/launch_with_deeplink'
```

migs647:
we're looking at this today

sanhue0302:
[...]

PR #954, opened 2020-10-27
Fix "Application Opened" event when app starts by deeplink

krin-san:
**What does this PR do?**
- Fixes assertion failures in Debug build when app starts by deeplink.
- Fixes missing "url" parameter in "Application Opened" tracking event

- Change app scheme to "Wait for the executable to be launched"
- Start app by deeplink
- Ensure app didn't crashed (without debugger) or stopped on assertion in SEGUtils.m file
- Check with Segment Debugger that "Application Opened" tracking event contains "url" parameter

https://developer.apple.com/documentation/uikit/uiapplicationlaunchoptionsurlkey?language=objc

#947

**Questions:**
  No.
  Most likely not as PR contains just fixes.
  Maybe? _That question is unclear and need to be rephrased_


krin-san:
That seems to be the wrong solution. `SEGFileStorage.dataFromJSON` now fails to serialize it. Would be safer to convert NSURL into a string right in `_applicationDidFinishLaunchingWithOptions`.

bsneed:
Closing this in lieu of #962

PR #956, opened 2020-10-28
Added category option to screen calls

bsneed:
Adds 3 new versions of the screen call to allow category to be passed while maintaining compatibility with previous call points.

```
- (void)screen:(NSString *)screenTitle category:(NSString * _Nullable)category properties:(SERIALIZABLE_DICT _Nullable)properties options:(SERIALIZABLE_DICT _Nullable)options;
- (void)screen:(NSString *)screenTitle properties:(SERIALIZABLE_DICT _Nullable)properties options:(SERIALIZABLE_DICT _Nullable)options;
- (void)screen:(NSString *)screenTitle category:(NSString * _Nullable)category properties:(SERIALIZABLE_DICT _Nullable)properties;
- (void)screen:(NSString *)screenTitle properties:(SERIALIZABLE_DICT _Nullable)properties;
- (void)screen:(NSString *)screenTitle category:(NSString * _Nullable)category;
- (void)screen:(NSString *)screenTitle;
```

### DIFF
--- a/Segment/Classes/SEGAnalytics.h
+++ b/Segment/Classes/SEGAnalytics.h
@@ -107,9 +107,13 @@ NS_SWIFT_NAME(Analytics)
  When a user views a screen in your app, you'll want to record that here. For some tools like Google Analytics and Flurry, screen views are treated specially, and are different from "events" kind of like "page views" on the web. For services that don't treat "screen views" specially, we map "screen" straight to "track" with the same parameters. For example, Mixpanel doesn't treat "screen views" any differently. So a call to "screen" will be tracked as a normal event in Mixpanel, but get sent to Google Analytics and Flurry as a "screen".
 
  */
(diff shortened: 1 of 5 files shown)


migs647 (approved):
LGTM

willemfrishert:
@bsneed unfortunately the change in SEGScreenPayload.h is a major change. Adding category to the initializer changes the signature on how to create a SEGScreenPayload.

Two questions/suggestions:
1. To keep it to a minor and backwards compatible, shouldn't we should duplicate the init and a new initializer SEGScreenPayload.h that takes in the category? (similar to your duplication of methods above?)
2. Should the initializer forces a nonnull values via `NS_ASSUME_NONNULL_BEGIN` & `NS_ASSUME_NONNULL_END`)? The `category`- and `properties `-properties that are available are marked to be nullable and readonly so they could possibly be initialized as such?

bsneed:
@willemfrishert thats a great call out, thanks!  Will add that shortly.

andrewtheis:
@bsneed If category can be nil, this needs proper nullable flag in order for the Swift interface to allow nil.

PR #957, opened 2020-10-28
Output a log message when events are trimmed from the queue

bsneed:
- Output a log message when events are trimmed from the queue.
- Replicated behavior of Android library.


migs647 (approved):
LGTM

Issue #752, opened 2018-02-10
Allow static library usage from Cocoapods


bridger:
Version [1.4.0 of Cocoapods](http://blog.cocoapods.org/CocoaPods-1.4.0/) allows a pod to be used as a static library, even in a Swift app. It requires the library's podspec to contain `s.static_framework = true`.

I believe this will help apps launch faster. Here's a blog post that explains why too many dynamic frameworks slows down launch: https://blog.automatic.com/how-we-cut-our-ios-apps-launch-time-in-half-with-this-one-cool-trick-7aca2011e2ea

f2prateek:
Thanks for the report! I've triaged this internally (https://segment.atlassian.net/browse/LIB-234). I don't have an ETA, but we'll keep this on our radar for upcoming sprints. PRs welcome in the meantime though!

krin-san:
It's a shame to see this issue lying around for already a year without any signs of work being done to solve it. [CocoaPods 1.5 release](http://blog.cocoapods.org/CocoaPods-1.5.0/) provides an easy solution for that issue. Both issue and solution are [well known to the community](https://github.com/ReactiveX/RxSwift/issues/1800#issuecomment-439187797).

Is there a reason not to include 1-line fix into the podspec file in 3.7.0 release?

For those who stumble upon the same issue and can't wait any longer, here a workaround:
1. For direct use in Podfile, just change the line like that:
```diff
- pod 'Analytics'
+ pod 'Analytics', :modular_headers => true
```
2. For inclusion in private library podspec:
- Download latest released `Analitycs.podspec` file
- Add this line inside:
```ruby
s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES' }
```
- Push modified podspec to your private Specs repo
- If library validation job runs on CI, ensure that `pod spec repo update` is executed to use your new shiny podspec
- In Podfile place your private Specs repo before the official one to ensure your podspec will be used

bsneed:
I'll see if I can get this in with some other changes today.  Apologies for the time taken on this.  Our library wasn't quite in shape to be able to do this until very recently.  Supporting Cocoapods, Carthage, and SPM on multiple platforms was extremely difficult before.

migs647:
We created a pull request: https://github.com/segmentio/analytics-ios/pull/968

Tested with Swift and Objective-C using `use_frameworks! :linkage => :static` and `use_modular_headers!` respectively.

PR #968, opened 2020-11-13
Added static lib support for cocoapods

migs647:
**What does this PR do?**
Added the capability for the pod to be built as a static framework

Test with Swift and Objective-C. It may be appropriate to use `use_modular_headers!` or `use_frameworks! :linkage => :static` in the Podfile

See above

Github issue: https://github.com/segmentio/analytics-ios/issues/752

https://segment.atlassian.net/browse/LIBMOBILE-211

**Questions:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Analytics is available through [CocoaPods](http://cocoapods.org) and [Carthage](
 ```ruby
 pod "Analytics", "3.7.0"
 ```
-Note: Dynamic frameworks are the default and recommended. If you prefer static libraries add `use_modular_headers!` or `use_frameworks! :linkage => :static` in your Podfile. 
+Note: Segment _strongly_ recommends that you use a dynamic framework to manage your project dependencies. If you prefer static libraries, you can add `use_modular_headers!` or `use_frameworks! :linkage => :static` in your Podfile. However, you must then _manually update_ all of your dependencies on a regular schedule.
 
 ### Carthage
 
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Analytics is available through [CocoaPods](http://cocoapods.org) and [Carthage](
 ```ruby
 pod "Analytics", "3.7.0"
 ```
+Note: Dynamic frameworks are the default and recommended. If you prefer static libraries add `use_modular_headers!` or `use_frameworks! :linkage => :static` in your Podfile. 
 
 ### Carthage
 
--- a/Analytics.podspec
+++ b/Analytics.podspec
@@ -17,7 +17,9 @@ Pod::Spec.new do |s|
 
   s.ios.deployment_target = '10.0'
   s.tvos.deployment_target = '10.0'
-  s.osx.deployment_target = '10.13'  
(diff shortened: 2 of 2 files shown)


bsneed (approved):
There should probably be additions to the readme about anything extra people might need to do or want to do to get static vs. dynamic.

bsneed (approved):

README.md
```diff
@@ -61,6 +61,7 @@ Analytics is available through [CocoaPods](http://cocoapods.org) and [Carthage](
 ```ruby
 pod "Analytics", "3.7.0"
 ```
+Note: Dynamic frameworks are the default and recommended. If you prefer static libraries add `use_modular_headers!` or `use_frameworks! :linkage => :static` in your Podfile.
```
sanscontext, line 64:
```suggestion
Note: Segment _strongly_ recommends that you use a dynamic framework to manage your project dependencies. If you prefer static libraries, you can add `use_modular_headers!` or `use_frameworks! :linkage => :static` in your Podfile. However, you must then _manually update_ all of your dependencies on a regular schedule.
```

fangmobile:
@sanscontext @bsneed 
We use Segment with corp license and use it as a dependency in our pod spec.
`.dependency 'Analytics', '~> 4.1'  `

with 
`s.static_framework = true`

our app that uses our internal library will get this error,
`target has transitive dependencies that include statically linked binaries: (Analytics)`

why are we making Segment as a static framework?

fangmobile:
@migs647 any idea to work around the error?

bsneed:
@fangmobile can you see if these options work for you?  https://guides.cocoapods.org/syntax/podfile.html#use_frameworks_bang

fangmobile:
@bsneed thanks `:linkage => :static` works in the app Podfile. 
We are also evaluating adding `s.static_framework = true` in our library pod spec. What are the pro and cons?

Cons: Since it is statically loaded App core memory usage is higher. App load time might be affected as well. There are some extra steps to do involving multiple copies of files produced by Cocoapods scripts.

Pros?

any thoughts are appreciated.

bsneed:
Being statically linked, the app load time will be shorter.  As for memory usage, that should be lower as well as there's no overhead of the frameworks.  If you're seeing a memory difference it could be because some frameworks are being lazy loaded, but if they are being used, they'll get loaded eventually.  Apple's had a few sessions that go over the benefits of static linking vs. dynamic, namely that they optimize the system such that performance is still good at, I think they said 4?? dynamic frameworks, and more than that it drops precipitously.  YMMV though.

This article covers some of it as well as links to the WWDC sessions on the subject:  https://medium.com/ios-expert-series-or-interview-series/ios-app-launch-static-binding-vs-dynamic-binding-linking-vs-embedded-e69ea9f03f72

I was thinking if you set the :linkage to be dynamic, it would override what we're setting, but not sure without checking.

Noobish1:
I'm seeing similar to @fangmobile where we have an internal CocoaPod and this change forces the Segment framework to be built as a static framework so we get transitive dependency errors.

Issue #972, opened 2020-11-26
Empty defaultSettings breaks event reporting


leethree:
We found there is major data discrepancy between our Android and iOS app using Segment and some iOS clients don't report any data to Segment. After investigation, we found that the library will not report any events if the first  call to `cdn-settings.segment.com` fails.

We are using `analytics-react-native` and it passes an empty dict for `defaultSettings`. 

https://github.com/segmentio/analytics-react-native/blob/9b5734d58b3f4099b5396504debb4ffad2abdd79/packages/core/src/configuration.ts#L29

This breaks the fallback logic of the library settings because here the code only checks for `nil`.

https://github.com/segmentio/analytics-ios/blob/9cab0239cf807e92402c48ea90ce516c92c7ec6a/Segment/Internal/SEGIntegrationsManager.m#L463-L470

`newSettings[@"integrations"][@"Segment.io"][@"apiKey"]` doesn't do anything because the keys don't exist. This results in an empty setting cached in the client and no events will be send to Segment.io because the integration is not enabled.

> No settings for Segment.io. Skipping.

bsneed:
Will have a look into this today.

bsneed:
Fixed in PR #973.  If any other details come up, please reopen this issue. Thanks so much for the report!!

PR #973, opened 2020-12-09
Data Residency Changes

bsneed:
- Adds support for data residency.
- New configuration constructor option to allow for a default API host to be supplied.
- New configuration property, `apiHost` to retrieve the host in use.
- ApiHost can be picked up from `/settings` endpoint and stored for future requests.
- Fixed issue where `defaultSettings` might be read and/or stored incorrectly.
- Fixed some project warnings in Xcode 12


bsneed:
This also fixes #972

migs647 (approved):
LGTM 👍🏻

PR #977, opened 2021-01-13
Pushed save of anonId higher up into the flow.

### LINKED ISSUE #974: Storing anonymous id after running middlewares
Hi,

In AdevintaSpain we are implementing several changes in our apps to be compliant with some privacy regulations. As a result we have implemented a middleware which anonymises some of the information sent depending on the user's privacy preferences. 

One of the values we are anonymising is the `anonymousId` sent on each event setting a constant value. This constant should only be sent if/while the user wants to anonymise it's data but what we are facing with the iOS SDK is that once we start anonymising the anonymous id, then we never see a proper anonymous id anymore even if the user changes its preferences.

# Steps to reproduce it
1. Create a middleware where you change the anonymous id.
2. Append the middleware to the `sourceMiddleware` array.
3. Send -at least- an identify event.
4. Remove the middleware and run the app again and send any event.

# What did you expect to happen?
In step 3 anonymous id should be modified, in step 4 it shouldn't.

# What happened instead?
After removing the middleware the anonymous id still was anonymised. After checking the SDK we've seen that you are inserting `SEGIntegrationsManager` in the last position of the middlewares array, and inside that middleware, when you process an identify event you do:

```objc
NSString *anonymousId = payload.anonymousId;
NSString *existingAnonymousId = self.cachedAnonymousId;
    
if (anonymousId == nil) {
    payload.anonymousId = anonymousId;
} else if (![anonymousId isEqualToString:existingAnonymousId]) {
    [self saveAnonymousId:anonymousId];
}
```

Please note several things here:
1. The first part of the condition does nothing: you are setting `payload.anonymousId` inside a variable, then checking if nil, and assigning to it its value. Not sure it make sense.
2. As we understand middlewares should not store any applied change but change the events on the fly for each specific case. Also notice that the Android SDK does not add this middleware nor stores this information. It is our understanding that both SDK should behave equally.

bsneed:
- Makes anonId save happen up in the identify() call now rather than in the guts of the system.
- Makes it such that middleware can modify the anonId without changing it for the entire system.
- Fixes #974

### DIFF
--- a/Segment/Classes/SEGAnalytics.m
+++ b/Segment/Classes/SEGAnalytics.m
@@ -274,6 +274,8 @@ NSString *const SEGBuildKeyV2 = @"SEGBuildKeyV2";
     NSString *anonId = [options objectForKey:@"anonymousId"];
     if (anonId == nil) {
         anonId = [self getAnonymousId];
+    } else {
+        [self.integrationsManager saveAnonymousId:anonId];
     }
     // configure traits to match what is seen on android.
     NSMutableDictionary *existingTraitsCopy = [[SEGState sharedInstance].userInfo.traits mutableCopy];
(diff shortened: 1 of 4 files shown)


migs647 (approved):
LGTM 👍🏻

miquelalvarado:
Works as expected! Thanks! 🙌

miquelalvarado (approved):

PR #982, opened 2021-01-27
Pull IDFA if it has changed and rebuild context if necessary.

bsneed:
- Fixes issue where IDFA is not updated until after a user has left the app and returned.
- Check IDFA values if ad-tracking enabled to determine if context needs to be rebuilt.


Segment/Classes/SEGAnalytics.m
```diff
@@ -27,6 +27,7 @@ @interface SEGAnalytics ()
 @property (nonatomic, strong) SEGStoreKitTracker *storeKitTracker;
 @property (nonatomic, strong) SEGIntegrationsManager *integrationsManager;
 @property (nonatomic, strong) SEGMiddlewareRunner *runner;
+@property (nonatomic, strong) NSString *lastIDFA;
```
migs647, line 30: Should this be copy instead of strong?

bsneed: yes, will fix on master.

Issue #975, opened 2021-01-04
SEGContext overwrites UIViewRepresentableContext


Paul-Svetlichny:
NS_SWIFT_NAME(Context)
@interface SEGContext : NSObject

This part in the header SEGContext.h overwrites SwiftUI UIViewRepresentable
typealias Context = UIViewRepresentableContext

Leads to application using SwiftUI and UIViewRepresetableContext as Context building fails.

migs647:
Thanks @Paul-Svetlichny -- we are looking into SwiftUI compatibility currently.

dmiluski:
Any update on this? Anybody find a work-a-round for the time being?

dmiluski:
@Paul-Svetlichny ,

With help from @bsneed , I was able to get Segment + SwiftUI Representable working. Hopefully this gets you moving forward too.

https://github.com/segmentio/analytics-ios/pull/985#issuecomment-776903974

```diff
import AuthenticationServices
import Segment
import SwiftUI
final class SignInWithApple: UIViewRepresentable {
-    func makeUIView(context: SignInWithApple.Context) -> ASAuthorizationAppleIDButton {
+    func makeUIView(context: SignInWithApple.Context) -> ASAuthorizationAppleIDButton {
        // 3
        return ASAuthorizationAppleIDButton()
    }
    // 4
-    func updateUIView(_ uiView: ASAuthorizationAppleIDButton, context: Context) {}
+    func updateUIView(_ uiView: ASAuthorizationAppleIDButton, context: SignInWithApple.Context) {}
}
```

bsneed:
Ah, didn't see this here, thanks @dmiluski !!

PR #985, opened 2021-02-10
Remove SwiftName/SwiftUI conflict

dmiluski:
Problem:
Segment Context Swift Name is conflicting with SwiftUI UIViewRepresentable

resolves https://github.com/segmentio/analytics-ios/issues/975

**What does this PR do?**

When using Segment + SwiftUI, there is a namespace collision between SegContext (using swift name Context) and SwiftUI Context, which isn't a part of SwiftUI namespace leading to a collision.

Attempt to integration with SwiftUI project that uses UIViewRepresentable/UIViewController Representable

example:
If using existing Segment branch, Context will conflict with Segment Context and fail to compile
```
import AuthenticationServices
import SwiftUI

final class SignInWithApple: UIViewRepresentable {

    func makeUIView(context: Context) -> ASAuthorizationAppleIDButton {
        // 3
        return ASAuthorizationAppleIDButton()
    }

    // 4
    func updateUIView(_ uiView: ASAuthorizationAppleIDButton, context: Context) {}
}
```

https://github.com/segmentio/analytics-ios/issues/975

**Questions:**

### DIFF
--- a/Segment/Classes/SEGContext.h
+++ b/Segment/Classes/SEGContext.h
@@ -48,7 +48,6 @@ typedef NS_ENUM(NSInteger, SEGEventType) {
 @protocol SEGMutableContext;
 
 
-NS_SWIFT_NAME(Context)
 @interface SEGContext : NSObject <NSCopying>
 
 // Loopback reference to the top level SEGAnalytics object.
--- a/SegmentTests/ContextTest.swift
+++ b/SegmentTests/ContextTest.swift
@@ -24,7 +24,7 @@ class ContextTests: XCTestCase {
         var exception: NSException?
         
         exception = objc_tryCatch {
-            context = Context()
+            context = SEGContext()
         }
         
         XCTAssertNil(context)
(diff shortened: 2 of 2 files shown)


dmiluski:
cc @migs647

Segment/Classes/SEGContext.h
```diff
@@ -48,7 +48,6 @@ typedef NS_ENUM(NSInteger, SEGEventType) {
 @protocol SEGMutableContext;
 
 
-NS_SWIFT_NAME(Context)
```
dmiluski, line 51: This could also be named something else, but this was the minimum change to resolve the naming conflict with SwiftUI

bsneed:
I can't merge this.  This is why swift have module namespaces.  Segment.Context will clear it up.  If we change class names (a breaking change for everyone) every time apple uses a common name, we might as well go back to prefixing like ObjC.

bsneed:
We looked at your example a little more, the fix for your code is this:

```
import AuthenticationServices
import Segment
import SwiftUI
final class SignInWithApple: UIViewRepresentable {
    func makeUIView(context: SignInWithApple.Context) -> ASAuthorizationAppleIDButton {
        // 3
        return ASAuthorizationAppleIDButton()
    }
    // 4
    func updateUIView(_ uiView: ASAuthorizationAppleIDButton, context: SignInWithApple.Context) {}
}
```

Classes that conform to UIViewRepresentable automatically get a type alias called `Context` which is what your conflict is stemming from.

dmiluski:
Thanks for the update

PR #993, opened 2021-04-23
Fixed issue where incorrect segment settings could be built

bsneed:
- Standardizes name for the segment destination into a const across the project.
- Makes building default settings object less complex.

### DIFF
--- a/Segment/Internal/SEGIntegrationsManager.m
+++ b/Segment/Internal/SEGIntegrationsManager.m
@@ -397,7 +397,7 @@ NSString *const kSEGCachedSettingsFilename = @"analytics.settings.v2.plist";
 - (void)updateIntegrationsWithSettings:(NSDictionary *)projectSettings
 {
     // see if we have a new segment API host and set it.
-    NSString *apiHost = projectSettings[@"Segment.io"][@"apiHost"];
+    NSString *apiHost = projectSettings[kSEGSegmentDestinationName][@"apiHost"];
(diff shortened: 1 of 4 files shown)


Segment/Internal/SEGIntegrationsManager.m
```diff
@@ -485,9 +496,9 @@ - (void)refreshSettings
                         NSMutableDictionary *newSettings = [self.configuration.defaultSettings serializableMutableDeepCopy];
                         NSMutableDictionary *integrations = newSettings[@"integrations"];
                         if (integrations != nil) {
-                            integrations[@"Segment.io"] = @{@"apiKey": self.configuration.writeKey, @"apiHost": [SEGUtils getAPIHost]};
+                            integrations[kSEGSegmentDestinationName] = [self segmentSettings];
                         } else {
-                            newSettings[@"integrations"] = @{@"integrations": @{@"apiKey": self.configuration.writeKey, @"apiHost": [SEGUtils getAPIHost]}};
```
migs647, line 490: So this would only happen if integrations never came down via settings? ie offline mode?

migs647 (approved):
LGTM

Segment/Internal/SEGIntegrationsManager.m
bsneed: yeah exactly.

nando0208 (approved):

PR #1009, opened 2021-07-30
Fix deadlock on first app launch

### LINKED ISSUE #1008: Deadlock on app launch
##  Description
Segment SDK can deadlock on app launch when using some of the integrations. Amplitude integration is one of those and I think it is one of the more popular ones.

The flaw in the integration manager module. It happens when integration initializer calls `dispatch_sync` with main queue. This is the case inside `-[Amplitude runSynchronouslyOnMainQueue:]` which is called from `-[Amplitude initializeApiKey:userId:setUserId:]`. This situation doesn't happen always because it is related to disk writing latency but in my app on some devices it was happening very often right after installation of the app from TestFlight. Detailed steps below.

*I will also provide PR fixing this*

## Prerequisities
- clean app install (no cached settings file)
- Amplitude integration configured
- very fast internet connection

## Steps
-> main queue
1. `-[SEGIntegrationsManager initWithAnalytics:]`
2. Configuration was downloaded very quickly and completion is handled on `io.segment.analytics` queue

-> `io.segment.analytics` queue
3. configuration written to file in `-[SEGIntegrationsManager setCachedSettings:]`
4. writing to file is finished but sometimes the code waits there and context gets switched
  - simulate it by putting [NSThread sleepForTimeInterval:1] after writing to file

-> main queue continues
5. `-[SEGIntegrationsManager onAppForeground:]`
6. `-[SEGIntegrationsManager refreshSettings]`
7. file exists on disk because already written in 3.
8. `-[SEGIntegrationsManager setCachedSettings:]`
9. Writing to file again takes time and context gets switched

-> `io.segment.analytics` queue continues
10. `-[SEGIntegrationsManager updateIntegrationsWithSettings:]`
11. Amplitude integration initializer calls dispatch_sync with main queue and waits

-> main queue
12. `-[SEGIntegrationsManager updateIntegrationsWithSettings:]`
13. calls dispatch_sync with `io.segment.analytics` queue and waits

FIN: deadlock because both `io.segment.analytics` and main queue are waiting for each other

## Example log
`[NSThread sleepForTimeInterval:2]` added after writing settings to file
```
00.522774 initWithAnalytics isMainThread 1
00.576702 refreshSettings isMainThread 1
00.929121 settingsRequestCompletionHandler isMainThread 0
00.929278 setCachedSettings isMainThread 0
01.080838 onAppForeground isMainThread 1
01.080872 refreshSettings isMainThread 1
01.080902 setCachedSettings isMainThread 1
02.941512 updateIntegrationsWithSettings isMainThread 0
03.007089 (Amplitude initializer) runSynchronouslyOnMainQueue isMainThread 0
03.085322 updateIntegrationsWithSettings isMainThread 1
(no further messages)
```

piotrtobolski:
**What does this PR do?**
Fixes #1008

`Segment/Internal/SEGIntegrationsManager.m`
it is small change

Requirements:
- App integrating segment SDK
- Fresh install of an app
- Amplitude integration configured
- Luck

This is race condition so it happens infrequently. To emulate problematic condition failing always before and not failing after the fix add `[NSThread sleepForTimeInterval:3];` inside `-[SEGIntegrationsManager setCachedSettings:]` line before `[self updateIntegrationsWithSettings:settings[@"integrations"]];`

See #1008

#1008

not UI/UX change

**Questions:**
  - No.
  - No.
  - I don't think so.

### DIFF
--- a/Segment/Internal/SEGIntegrationsManager.m
+++ b/Segment/Internal/SEGIntegrationsManager.m
@@ -156,7 +156,9 @@ NSString *const kSEGCachedSettingsFilename = @"analytics.settings.v2.plist";
 
 - (void)onAppForeground:(NSNotification *)note
 {
-    [self refreshSettings];
+    seg_dispatch_specific_async(_serialQueue, ^{
+        [self refreshSettings];
+    });
 }
 
 - (void)handleAppStateNotification:(NSString *)notificationName


bsneed:
Relaying from issue #1008 

Thanks for reporting this; however I don't think a sleep is the proper fix here.  I'm unable to reproduce locally, however from your triage of the issue, the fix *should* be that amplitude SDK should stop doing dispatch_sync on the main queue when it is called from the main queue.  Or optionally, the amplitude segment integration should call Amplitude.init from a background thread if that is required.  

Typically, using a sleep to fix a race condition just pushes said race condition elsewhere.

piotrtobolski:
Copied from issue #1008

@bsneed please revise and reopen the issue. I think there is some misunderstanding. This is just issue report, not a solution. Sleep is used to simulate writing to disk taking long time. It isn't used for the fix but to ensure whoever is debugging it has 100% reproduction.

If you check the source changes in attached PR #1009 (It's just 3 lines) you will see that the fix is to ensure that `-[SEGIntegrationsManager refreshSettings]` is always called on the same (segment private) queue ensuring that shared resource access is synchronized. This solution isn't moving the race condition elsewhere but removes it completely.

I considered making a fix in Amplitude but I decided it isn't the culprit here. It is calling dispatch_sync with main only when code is executed from non-main queue and it is required to call UIKit methods. Segment SDK is the one that is calling dispatch_sync with private queue from main and blocking it. This PR fixes this. This is result of my investigation. Of course, there may be other solutions (on integrated SDKs side as well) but I think it is still worth implementing this fix because integrating with other SDKs is Segment core feature so it is important to be a good citizen and avoid blocking the main queue.

bsneed (approved):

Issue #967, opened 2020-11-12
Braze Integration - AppBoy object of "Integrations" is always false


danielhorv:
Hi Guys!
We already have a functional Segment tracking in our iOS app, but we also need to integrate Braze (https://github.com/Appboy/appboy-segment-ios).
We have a problem, when the events are tracked, they include an "integrations" object with "Appboy" which is always false:

```
,"integrations": {
  "Appboy": false
},
```

This means, that Segment doesn't send the tracked events to Braze. I just looked into the source code and it will always set to false. Is there any reason why? 

```objective-c
- (NSDictionary *)integrationsDictionary:(NSDictionary *)integrations
{
    NSMutableDictionary *dict = [integrations ?: @{} mutableCopy];
    for (NSString *integration in self.analytics.bundledIntegrations) {
        // Don't record Segment.io in the dictionary. It is always enabled.
        if ([integration isEqualToString:@"Segment.io"]) {
            continue;
        }
        dict[integration] = @NO;
    }
    return [dict copy];
}
```

I just changed `dict[integration]` to `@YES` and now it seems to work.
Android just send an empty object and it's also working.

BR
Daniel

bsneed:
@danielhorv is your intent for braze to be delivered via cloud mode, or via device mode?

danielhorv:
I think we need both because we want to see all our events in Segment, but also want to use Braze for the push notifications and in-app messaging. Which requires some extra configuration for Braze.

Without the Appboy-Segment SDK, we receive every event out of the box in our Braze dashboard (destination is configured in Segment) - It's the cloud mode if I'm right. 

But we need the Appboy-Segment SDK (because of the push registration) and then `"Appboy": false` will be sent in every Analytics track/screen event. In this case, we don't receive any events on the Braze dashboard.

alanjcharles:
Closing since there has been no response in 10 months. Please feel free to reopen with additional information if this is still an issue.

PR #1027, opened 2022-01-24
Enable bundled integrations if specified

matibzurovski:
**What does this PR do?**
This PR fixes an issue where events wouldn't be delivered to integrations if their SDKs are bundled. Even when following the SDK [docs](https://segment.com/docs/connections/sources/catalog/libraries/mobile/ios/#selecting-destinations), integrations would be overridden and set to `false`. 

1. Integrate `analytics-ios` and [Braze](https://github.com/Appboy/appboy-segment-ios) into the same project. 
2. Set up Segment's `AnalyticConfiguration` to `use(SEGAppboyIntegrationFactory.instance())`
3. Track an event and set the options to allow Braze:
```swift
Analytics.shared().track("example", properties: nil, options: [
    "integrations": [
        "Appboy": true
    ]
])
```

4. Notice how the event JSON sent to the SDK has Appboy set to false, when it should actually be true.
```JSON
"integrations": {
    "Appboy": false
}
```

Issue was reported on https://github.com/segmentio/analytics-ios/issues/967.

**Questions:**
No

No

Not sure

### DIFF
--- a/Segment/Classes/SEGSegmentIntegration.m
+++ b/Segment/Classes/SEGSegmentIntegration.m
@@ -263,7 +263,10 @@ NSUInteger const kSEGBackgroundTaskInvalid = 0;
         if ([integration isEqualToString:kSEGSegmentDestinationName]) {
             continue;
         }
-        dict[integration] = @NO;
+        if ([dict objectForKey:integration] == nil) {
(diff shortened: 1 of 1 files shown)


matibzurovski:
thoughts @prayansh ?

PR #1070, opened 2026-03-31
Fix private header import in SEGReachability

### LINKED ISSUE #1069: Error "segment Use of private header from outside its module: 'netinet6/in6.h'"
After updating the Xcode to v26.4. The segment library is thrown an error.   

The reachability file accessing the private header 
https://github.com/segmentio/analytics-ios/blob/4b0968947eb916b3a29d56bd5e17d620ff29a220/Segment/Classes/SEGReachability.m#L26

Please suggest the solution for this ?

For now I am downgrading my Xcode version to unblock.

natanrolnik:
`<netinet/in.h>` already provides the public IPv6 types needed here, so this avoids the Xcode 26.4 module violation (Use of private header from outside its module)

Fixes #1069


aadeshAvoma (approved):

thedeveloperr:
@aadeshAvoma @natanrolnik  When can we merge this? Which versions will contain this fix?

